PR #140, opened 2019-07-20
Add French Translation

Arias800:
Maybe he can rest some spelling error but I tried to make the more clean translation possible ^^


lolamtisch:
Sorry looks like i clicked on the wrong button before

PR #164, opened 2019-08-06
Add NekoSama

Arias800:
Requested here : https://github.com/lolamtisch/MALSync/issues/163


Arias800:
It's fix but I'm not sure if I have used `utils.getBaseText` correctly

Issue #188, opened 2019-09-13
Episode detection on Kissanime


DaveDuck321:
**Describe the bug**
Episode detection on Kissanime does not work when the Episode URL looks like Episode-##, this pattern is always detected as Episode 1.

**Steps to reproduce the behavior:**
Psycho Pass is an example of this; here the episode tracking and detection will not work. I was able to reproduce this on both Chrome and Firefox.

**Additional information:**
 - Browser: Chrome & Firefox
 - Type: Extension
 - Url: https://kissanime.ru/Anime/Psycho-Pass

DaveDuck321:
This resolves my issue. Thank you.

PR #189, opened 2019-09-13
[BUGFIX] Kissanime episode identification

DaveDuck321:
Fixes issue #188 
This should only change the behavior when episode detection fails, else it will behave identically to the old system. This is better than the solution I asked about on discord.


lolamtisch:
Found a better solution. Check the issue for that

PR #307, opened 2020-02-17
AnimeZone fix

VATICAN-PSYCHO:
Recently, AnimeZone changed some urls. The site currently requires login to view content (episode mirror list).

### DIFF
--- a/src/pages/pageUrls.js
+++ b/src/pages/pageUrls.js
@@ -322,7 +322,7 @@ module.exports = {
   AnimeZone: {
     match: [
       '*://www.animezone.pl/odcinki/*',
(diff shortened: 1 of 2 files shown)


src/pages/pageUrls.js
```diff
@@ -322,7 +322,7 @@ module.exports = {
   AnimeZone: {
     match: [
       '*://www.animezone.pl/odcinki/*',
-      '*://www.animezone.pl/odcinek/*'
+      '*://www.animezone.pl/anime/*'
```
lolamtisch: The old url schema with odcinek still works. It is better to keep the old url in.

PR #373, opened 2020-05-16
Fill missing types and refactor

seahindeniz:
I have made some changes on pageInterface.ts to have type definitions and rest is refactoring regarding the types


seahindeniz:
That's odd, old commits are here

PR #397, opened 2020-06-20
add Animixplay

henrik9999:


src/pages/AniMixPlay/tests.json
```diff
@@ -0,0 +1,28 @@
+{
+  "title": "AniMixPlay",
+  "url": "https://animixplay.com/",
+  "testCases": [
+    {
```
watzitoya, line 5: missing v1 and v4
https://animixplay.com/v1/kaguya-sama-wa-kokurasetai-tensai-tachi-no-renai-zunousen-2
https://animixplay.com/v4/4-kaguya-sama-wa-kokurasetai-tensai-tachi-no-renai-zunousen-2nd-season

src/pages/AniMixPlay/main.ts
```diff
@@ -0,0 +1,55 @@
...
+  type: 'anime',
+  isSyncPage(url) {
+    return true;
+  },
+  sync: {
+    getTitle(url) {
+      return j.$('span.animetitle').text();
+    },
+    getIdentifier(url) {
+      return url.split('/')[4];
```
watzitoya, line 15: url with /v2/, the slug located at slash [5]

src/pages/AniMixPlay/tests.json
henrik9999: Hello,
I knew of the slug do version 2 being at another part but I considered this not being a problem because the number before that on v2 also seemed to be unique since you can just open the anime with only the number.
I didn't add the v1 test because the V1 url somehow can not keep the selected episode and defaults to no episode selected.
About v4 I didn't know about that one.
And lastly what do you mean with different page handle.

watzitoya: oh, just for unique id. 
yeah that number is unique, v4 also unique.
there are multiple different page for each anime, i thought need to test all variant.

henrik9999:
> need different page type handle

I am still not sure what you mean with that

PR #471, opened 2020-09-22
add vivo iframe on anime4you

henrik9999:


henrik9999:
Will change this code to something more trustworthy later

PR #495, opened 2020-10-12
add vidstreaming.us 

riora123:


henrik9999:
This is not correct, also I already fixed it in my pr

PR #552, opened 2021-01-03
new Url

ScarVite:
The Site changed the domain a while back


lolamtisch:
That is an loadbalancer domain addition. Not really needed to change that as you will be redirected automatically.

PR #590, opened 2021-02-06
Lint files, refactor

seahindeniz:


src/pages/Emby/main.ts
```diff
@@ -341,7 +341,7 @@ export const Emby: pageInterface = {
     testApi()
       .catch(() => {
         con.info('Not Authenticated');
-        return askForApiKey();
+        askForApiKey();
```
lolamtisch, line 344: askForApiKey returns an promise. return is needed here or atleast await

lolamtisch (changes requested):

PR #601, opened 2021-02-11
Update animestreamingfr.fr

TriForMine:
Update animestreamingfr.fr detection to handle the new version of it that uses next.js.


TriForMine:
This pull request is now ready since the new version of animestreamingfr.fr got released.

PR #623, opened 2021-03-02
Add MangaHub

MrSuperKing143:


src/pages/MangaHub/main.ts
```diff
@@ -0,0 +1,142 @@
...
+      elementEp(selector) {
+        return MangaHub.sync.getEpisode(MangaHub.overview!.list!.elementUrl(selector));
+      },
+    },
+  },
+  init(page) {
+    api.storage.addStyle(require('!to-string-loader!css-loader!less-loader!./style.less').toString());
+    j.$(document).ready(function() {
+      utils.urlChangeDetect(() => {
+        utils.waitUntilTrue(
```
lolamtisch, line 85: I dont think this is the best solution, as there is a wait afterward too. 

You could detect with the url if it is an overview/sync page this way you could save the wait and dont need to check for the #mangareader.

Sometimes this make problems as the content of the previos page is still loaded during the url change. An alternative to urlChangeDetect is to use ChangeDetect(callback, checkFunction). It will trigger if the return of the checkFunction is different. You for example return the title or something that changes when the page is loaded.
https://github.com/MALSync/MALSync/blob/e8125839ee0e92a3f7af806342f2da626c67663d/src/pages/Twistmoe/main.ts#L103

src/pages/MangaHub/main.ts
```diff
@@ -0,0 +1,142 @@
...
+          function() {
+            return !j.$('html.no-js.nprogress-busy').length;
+          },
+          function() {
+            page.reset();
+            if (j.$('#mangareader').length) {
+              utils.waitUntilTrue(
```
lolamtisch, line 92: This could result in triggering multiple times at the same time. The function returns a interval that could be destroyed before setting an new wait.
https://github.com/MALSync/MALSync/blob/e8125839ee0e92a3f7af806342f2da626c67663d/src/pages/DreamAnimes/main.ts#L58-L59

This and the comment before could be fixed for the popup of multiple UI on the overview of an anime.
Ex.:
open https://mangahub.io/manga/no-game-no-life
Click on one "Popular Manga Updates" on the bottom of the page. Now the ui should show multiple times.
image

lolamtisch (changes requested):

lolamtisch (approved):

PR #676, opened 2021-04-14
Update Kangaryu style

NatsuDzn:
They updated their design recently so the container isn't matching with the whole website.
Fixed in this PR.


lolamtisch (approved):

lolamtisch:
Thank you

PR #745, opened 2021-05-31
add Thailand translation

F1rstStr0ke:


lolamtisch:
Next time you could use the helper tool
https://lolamtisch.github.io/MALSync-localization

PR #818, opened 2021-08-15
Anistream integration

Flumuffel:


pages.md
```diff
@@ -13,6 +13,13 @@
     </thead>
```
lolamtisch, line 13: This list is autogenerated. Please remove this from the Pull Request.

src/pages/list.json
```diff
@@ -59,6 +59,11 @@
     "type": "anime",
```
lolamtisch, line 59: This list is autogenerated. Please remove this from the Pull Request.

test/headless/test.js
```diff
@@ -252,7 +252,7 @@ async function testPageCase(block, testPage, page) {
 }
 
 async function loopEl(testPage, headless = true) {
-  // if(testPage.title !== 'Kissmanga') return;
```
lolamtisch, line 255: Remove this change too.

src/pages/Anistream/main.ts
```diff
@@ -0,0 +1,83 @@
+import { pageInterface } from '../pageInterface';
+
+let jsonData;
+
+export const Anistream: pageInterface = {
+  name: 'Anistream',
+  domain: 'https://Anistream.de',
+  languages: ['German'],
+  type: 'anime',
+  database: 'Anistream',
```
lolamtisch, line 10: Please remove the database key, as it is only for first party implementations.

src/pages/Anistream/main.ts
```diff
@@ -0,0 +1,83 @@
...
+    },
+  },
+  init(page) {
+    api.storage.addStyle(require('!to-string-loader!css-loader!less-loader!./style.less').toString());
+
+    let oldJson;
+
+    j.$(document).ready(function () {
+      setInterval(function () {
+        utils.waitUntilTrue(
```
lolamtisch, line 64: Please use `utils.changeDetect` instead of the interval and waitUntilTrue combo. The first function is called when a change is detected and the second one is the data that should be watched (just pass the json text without decoding here).

Example: https://github.com/MALSync/MALSync/blob/master/src/pages/Simplyaweeb/main.ts#L36

lolamtisch (changes requested):
On top of the comments above there is no json on the overview page atm.

lolamtisch (changes requested):
The Id on the overiview and video page are still different.

PR #971, opened 2021-11-10
[TASK] Chnage animesuge TLD from .io to .to

amitojsingh366:
The owner of animesuge has switched domains from [animesuge.io](https://animesuge.io) to [animesuge.to](https://animesuge.to), and this made MALSync stop working.
These changes should fix the above mentioned issue

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,7 +270,7 @@ Makes it possible to use your MyAnimeList/Anilist/Kitsu/Simkl anime/mangalist as
                 <td><a href="https://dynasty-scans.com"><img src="https://favicon.malsync.moe/?domain=https://dynasty-scans.com"> DynastyScans</a></td>
                 <td></td>
               </tr><tr>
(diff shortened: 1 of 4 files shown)


henrik9999:
its already fixed for the next release

amitojsingh366:
oh ok, when is it getting released?

henrik9999:
there are no eta

but you can use a workaround in malsync for now

You will need to add the custom domain like below. There is a button Custom Domains in the settings. You possibly need to restart the browser afterwards.

```animesuge``` -> ```https://animesuge.to/```

or you could just use 9anime since animesuge is just using 9anime

amitojsingh366:
> there are no eta
> 
> but you can use a workaround in malsync for now
> 
> You will need to add the custom domain like below. There is a button Custom Domains in the settings. You possibly need to restart the browser afterwards.
> 
> `animesuge` -> `https://animesuge.to/`
> 
> or you could just use 9anime since animesuge is just using 9anime

thanks for your help, it worked!

PR #1001, opened 2021-12-20
Make spacing consistent

PythonCoderAS:
The spacing was off for some lines and it was causing a minor annoyance. 

I potentially recommend using something like prettier to make the entire project consistent.


lolamtisch (approved):

PythonCoderAS:
> We use eslint+prettier, but only for project files. No one really cares about the few build js files.

Shouldn't be too hard to set it up for the build js as well, especially prettier. I'll PR in a new config soon.

PR #1019, opened 2022-01-10
Fixed bug regarding manga volumes in discord rich presence

Gamesmes90:
I found a bug regarding the rich presence. 

I was testing the extension with Komga and found out that if the manga has "volume" or "volumes" tag, the rich presence will display "Chapter: 0 of X".

image

Checking for the volumes fixes the issue. I also called the "UI_Volume" directly.

image

### DIFF
--- a/src/pages/syncPage.ts
+++ b/src/pages/syncPage.ts
@@ -1194,16 +1194,20 @@ export class SyncPage {
             }
           }
           if (typeof this.curState.episode !== 'undefined') {
-            let totalEp;
-            if (this.curState.volume > 0){
-              const ep = this.curState.volume;
-              totalEp = this.singleObj.getTotalVolumes();
-              if (!totalEp) totalEp = '?';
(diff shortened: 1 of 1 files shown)


henrik9999 (changes requested):
There are pages that have volume and chapter at the same time I would prefer that it shows the chapter number or maybe both instead then

Gamesmes90:
> There are pages that have volume and chapter at the same time I would prefer that it shows the chapter number or maybe both instead then

Sorry, I didn't know that. I can try adding both. What pages can show both?

Gamesmes90:
Now it shows both volume and chapter number.

image

I changed "of" with a slash to save space when it shows both.

lolamtisch:
I think the chapter should be in front, as it is the more important metric. Would be annoying if it would cut of.

Gamesmes90:
Done.

image

lolamtisch:
I cleaned up the code. 
I removed the untranslated "of". 
And the totalchap/volume is only shown if there is one as the ? wasnt that helpfull in the presence

PR #1020, opened 2022-01-10
Update Flame Scans

MrSuperKing143:


lolamtisch (changes requested):
Seems like the identifier are different on the overview page and the chapter page.

https://flamescans.org/1642392061-omniscient-readers-viewpoint-chapter-90/
https://flamescans.org/series/1642392121-omniscient-readers-viewpoint/

PR #1043, opened 2022-02-04
Update AnimeOnsen

mist8kengas:
*remaking pull request due to deleting old fork


henrik9999:
please add tests aswell (test.json)

mist8kengas:
Tests Prettier says to format with newlines, VSCode says not to and automatically formats to remove newlines.
image
image

lolamtisch:
grafik
Can not reproduce it. Just fix it so that `build (pull_request)` is fine

src/pages/AnimeOnsen/main.ts
```diff
@@ -2,70 +2,111 @@ import { pageInterface } from '../pageInterface';
...
-      const identifier = urlParams.get('v');
-      return identifier || '';
+      // get animeonsen content id for database
+      const [, , contentId] = new URL(url).pathname.split('/');
+      return contentId || '';
     },
     getOverviewUrl(url) {
       // generate ao.details url
-      const urlParams = new URL(url).searchParams;
+      const [, , contentId] = new URL(url).pathname.split('/');
```
lolamtisch, line 69: The contentId is retrived 3 times exactly the same way. I would recommend to just call the overview.getIdentifier directly.
`AnimeOnsen.overview!.getIdentifier(url);`

src/pages/AnimeOnsen/main.ts
```diff
@@ -74,15 +115,22 @@ export const AnimeOnsen: pageInterface = {
       // handlePage()
       page.handlePage();
 
-      // go-home + episode-next
-      j.$('div#ao-episode-buttons-wrapper').on('click', () => {
+      // handle going back
+      j.$('div.ao-player-metadata-navigation button').on('click', () => {
+        setTimeout(() => {
+          page.handleList();
```
lolamtisch, line 121: handleList is used to update the episode List if for example the page has a pagination. What exactly was the idea behind this. Did I miss something?

src/pages/AnimeOnsen/style.less
```diff
@@ -0,0 +1,32 @@
+@import './../pages';
```
lolamtisch, line 1: The styling for mal-sync-active is missing. It is added on the episode list signalising the last watched episode or watched episodes. Normaly it changes the backgroundcolor of the element. But outline would be a possibility here too.
image

src/pages/AnimeOnsen/main.ts
```diff
@@ -2,70 +2,111 @@ import { pageInterface } from '../pageInterface';
...
-      return identifier || '';
+      // get animeonsen content id for database
+      const [, , contentId] = new URL(url).pathname.split('/');
+      return contentId || '';
     },
     getOverviewUrl(url) {
       // generate ao.details url
-      const urlParams = new URL(url).searchParams;
+      const [, , contentId] = new URL(url).pathname.split('/');
       const overviewUrl = new URL(`https://animeonsen.xyz`);
```
lolamtisch, line 70: Better to use AnimeOnsen.domain here too.

lolamtisch (changes requested):

src/pages/AnimeOnsen/main.ts
mist8kengas: in the UI, this is the burger button to go back to the details page, it used to be the previous episode button on an old version of the site.

lolamtisch: It is not needed if there is a page load in between. It is just for the case you have a episode pagination and you switch the pagination page. Means in case the overview page stays but the episode list changes afterwards which is here not the case

lolamtisch (approved):

PR #1057, opened 2022-02-24
Adding animetoast

snoweuph:


src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,82 @@
...
+        return $('.entry-title.light-title').text() != null && url.includes('?link=');
+    },
+    isOverviewPage: function (url: string): boolean {
+        return $('.entry-title.light-title').text() != null && !url.includes('?link=');
+    },
+    sync: {
+        getTitle: function (url: string): string {
+            let title = $('.entry-title.light-title').text().split(' ');
+            if(!(title.length >= 2))return title.join(' ');
+            for(let i = title.length - 1; i >= title.length - 2; i--){
```
lolamtisch, line 19: Better use a regex that removes the parts you dont want with replace then you dont need all this split and loop logic

src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,82 @@
...
+                const t = title[i];
+                t.match(/(Ger|Sub|Dub)/gi) ? title.splice(i, 1) : null;
+            }
+            return title.join(' ');
+        },
+        getIdentifier: function (url: string): string {
```
lolamtisch, line 25: I would recomment to use the url path as identifier
animetoast.cc/dagashi-kashi-2-ger-sub/ -> dagashi-kashi-2-ger-sub

src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,82 @@
...
+            let title = $('.entry-title.light-title').text().split(' ');
+            if(!(title.length >= 2))return title.join(' ');
+            for(let i = title.length - 1; i >= title.length - 2; i--){
+                const t = title[i];
+                t.match(/(Ger|Sub|Dub)/gi) ? title.splice(i, 1) : null;
+            }
+            return title.join(' ');
+        },
+        getOverviewUrl: function (url: string): string {
+            return url.split('?link=')[0];
```
lolamtisch, line 35: Breaks on https://www.animetoast.cc/dagashi-kashi-2-ger-sub/?referrer=android&link=4

src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,82 @@
...
+        },
+        getEpisode: function (url: string): number {
+            let s = $('.current-link').text().match(/(\d+)/gi);
+            return Number((s![s!?.length-1]));
```
lolamtisch, line 39: I think you can remove the `g` from the regex flags then the index of the array stays the same.
using ! is here the wrong way. Better check if s is null and return a NaN. And better use real variable names not only s.
```
if(!s) return NaN;
return Number(s[2]);
```

src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,82 @@
...
+            offsetHandler: true,
+            elementsSelector: function(): JQuery<HTMLElement> {
+                return $('.tab-content > div > a');
+            },
+            elementUrl: function (selector: JQuery<HTMLElement>): string {
+                return selector.attr('href')!;
+            },
+            elementEp: function (selector: JQuery<HTMLElement>): number {
+                let s = selector.text().match(/(\d+)/gi);
+                return Number((s![s!?.length-1]));
```
lolamtisch, line 69: Same code as above. DRY (Don't repeat yourself). I recommend a seperate function that takes the text as input and returns the number.

lolamtisch (changes requested):

src/pages/Animetoast/main.ts
snoweuph: there are also subpages that are not animes like https://www.animetoast.cc/wochenplan-3/ is that still okay ?

lolamtisch: On these pages it should not even execute. You can do that over `isOverviewPage` and `isSyncPage`, if both are false then it wont do anything. Normaly we check for special elements on the page that are only on the overview page like the episode list or something like that

snoweuph: ahh okay, good, did it in the fixed version how you suggested

src/pages/Animetoast/main.ts
snoweuph: I made a extra function how you suggested

snoweuph:
Could you please take a extra look onto my Regex expressions, if they could be improved?

src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,75 @@
...
+  },
+  sync: {
+    getTitle(url: string): string {
+      return $('.entry-title.light-title')
+        .text()
+        .replace(/(Ger Sub|Ger Dub)/gi, '')
+        .trim();
+    },
+    getIdentifier(url: string): string {
+      return url.replace(/(https|http):\/\//gi, '').split('/')[1];
```
lolamtisch, line 23: You can just use our utils funcition here. `utils.urlPart(url, 3)`

src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,75 @@
...
+    },
+    getOverviewUrl(url: string): string {
+      return `${Animetoast.domain}/${Animetoast.sync.getIdentifier(url)}`;
+    },
+    getEpisode(url: string): number {
+      const matching = $('.current-link').text().match(/(\d+)/i);
+      return matching ? parseEpisode(matching) : NaN;
```
lolamtisch, line 30: please move the regex and condition into the paresEpisode. Just the $('.current-link').text() should be passed to the function.

src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,75 @@
...
+      return Animetoast.sync.getTitle(url);
+    },
+    getIdentifier(url: string): string {
+      return Animetoast.sync.getIdentifier(url);
+    },
+    uiSelector(selector: string) {
+      Animetoast.sync.uiSelector!(selector);
+    },
+    list: {
+      offsetHandler: true,
```
lolamtisch, line 50: Not needed, can be removed. As far as I see the eps start by 1.

src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,75 @@
...
+      elementsSelector(): JQuery<HTMLElement> {
+        return $('.tab-content > div > a');
+      },
+      elementUrl(selector: JQuery<HTMLElement>): string {
+        return selector.attr('href')!;
+      },
+      elementEp(selector: JQuery<HTMLElement>): number {
+        return parseEpisode(selector);
```
lolamtisch, line 58: Same as above. And it is wrong either way. Here probably selector.text() is the way to go

src/pages/Animetoast/main.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const Animetoast: pageInterface = {
+  name: 'Animetoast',
+  domain: 'https://www.animetoast.cc',
+  languages: ['German'],
+  type: 'anime',
+  isSyncPage(url: string): boolean {
+    return $('.entry-title.light-title').text() != null && url.includes('link=');
+  },
+  isOverviewPage(url: string): boolean {
```
lolamtisch, line 12: Is true on pages like https://www.animetoast.cc/a-z-index/
Checking if the class .single-post exist is probably enough here.

lolamtisch (changes requested):

src/pages/Animetoast/tests.json
```diff
@@ -0,0 +1,31 @@
...
+          "title": "Vinland Saga",
+          "identifier": "vinland-saga-ger-sub",
+          "overviewUrl": "https://www.animetoast.cc/vinland-saga-ger-sub",
+          "episode": 10,
+          "nextEpUrl": "https://www.animetoast.cc/vinland-saga-ger-sub/?link=82",
+          "uiSelector": true
+        }
+      },
+      {
+        "url": "https://www.animetoast.cc/one-piece/?link=163",
```
lolamtisch, line 18: Two times a test for the video page and not for the overview. 
Example:
https://github.com/MALSync/MALSync/blob/master/src/pages/Gogoanime/tests.json

lolamtisch (changes requested):

src/pages/Animetoast/main.ts
snoweuph: Not allways https://www.animetoast.cc/one-piece/

src/pages/Animetoast/main.ts
snoweuph: would you do it like this?
```ts
isSyncPage(url: string): boolean {
    return $('.single-post') != null && url.includes('link=');
  },
  isOverviewPage(url: string): boolean {
    return $('.single-post') != null && !url.includes('link=');
  },
  ```

snoweuph:
are my Regex expressions okay, or can they be improved?

src/pages/Animetoast/main.ts
snoweuph: *pushed it allredy as change*

snoweuph:
After Changes Start Page gets seen as a Sync Page
image
image
image

snoweuph:
## Look After Improving Css
---
### Not in List
image
---
### Default Settings
image
---
### Highlight all Watched Episodes on
image
---
### Selected Episode + Highligted
image

snoweuph (commented):
For One Piece the Episode Number is Changing Around a bit on Animetoast, its not stable enough to be used for the test

snoweuph:
its now 20 days later, I would love to get a update to the current state, I allready requested a re-review :)

PR #1099, opened 2022-04-19
pagefixing

henrik9999:


MulverineX:
> The original locations for the id are gone. The new simpler location is present in this way in both the overview & playback pages, added by the site admin on request 😄.

PR #1105, opened 2022-04-25
Update animepahe id location

MulverineX:
The original locations for the id are gone. The new simpler location is present in this way in both the overview & playback pages, added by the site admin on request 😄.


henrik9999:
It's already fixed in #1099😂

PR #1107, opened 2022-04-26
Fix Cubari proxy chapter recognition

hotsno:
Currently the code for Guya/Cubari just looks at the URL to get the chapter number, but this is inaccurate for Cubari most of the time because of extra chapters being counted as regular chapters (ex. [if Chapter 11.5 is the first extra chapter, the URL for it will be 12](https://cubari.moe/read/mangasee/Spy-X-Family/)).

Specifically for the MangaSee (but possibly other) proxy, this can be fixed by looking at the chapter title, which can be found in the HTML, instead of the URL. After finding the chapter title, we can extract the accurate chapter number from there.

I made a new Cubari source and removed Cubari support from the Guya page, but I don't know if this is better or not.

### DIFF
--- a/src/pages/Cubari/main.ts
+++ b/src/pages/Cubari/main.ts
@@ -41,14 +41,17 @@ export const Cubari: pageInterface = {
       );
     },
     getEpisode(url) {
-      let temp = j
-        .$(
-          '#rdr-main > aside > div.rdr-aside-content > section.rdr-selector > div.rdr-selector-mid > div.rdr-chap-wrap.UI.FauxDrop > label',
-        )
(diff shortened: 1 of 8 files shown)


lolamtisch (changes requested):

akulakovsky:
Currently there's a lot of edge cases that need to be considered to make this work well, and Cubari v2 will probably handle almost all of these so it's probably better to wait till then.

hotsno:
Cubari v2 will likely be able to properly map MangaSee chapters, so it's easier to just wait till then.

PR #1116, opened 2022-05-13
Add missing video domain for voe player

louisroyer:
This domain is currently used for voe player on voiranime.com.


henrik9999:
please run `npm run build:missingPermissions` and commit what it generates

lolamtisch (changes requested):
^^

PR #1123, opened 2022-05-17
Added short Skip Forward/Backward shortcut to Player

Senjar:
Added a short Skip Forward/Backward shortcut with separate time intervals.
Default Skip Forward is 30sec and Skip Backward is 10sec (Configurable from settings). 
The shortcuts are not preconfigured to avoid confusion with old users (any default shortcut suggestions?).
It follows the same logic as Skip Intro.

### DIFF
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -854,6 +854,34 @@
       }
     }
   },
+  "settings_Shortcuts_Short_Skip_Forward": {
+    "message": "Skips forward the video",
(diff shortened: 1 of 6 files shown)


Senjar:
This makes the experience more consistent with all the players. Many players don't have skip buttons and the shortcuts some players have differ from site to site. But of course you may not want to add another shortcut.

lolamtisch:
I am going to close this. As we are decided against it. But still thanks for the pull request 👍

PR #1135, opened 2022-05-27
Addition of Luminous Scans

mac5617:


henrik9999:
If this is the exact same code as realmscans it might be better to do something like this
https://github.com/MALSync/MALSync/blob/master/src/pages/EdelgardeScans/main.ts#L4
https://github.com/MALSync/MALSync/blob/master/src/pages/ZeroScans/main.ts#L5
We should really add support for themes like on tachiyomi or something

mac5617:
Sure, I can change it

mac5617:
What do you do when the original website shuts down or changes format? For example, zero scans changed their website a month or two ago.

henrik9999:
> What do you do when the original website shuts down or changes format? For example, zero scans changed their website a month or two ago and it seems like there are a few websites that link to it.

well you can see in #1137 what happens another site will just do it like i said its not really optimal and something like on tachiyomi would be better but someone has to do it

PR #1141, opened 2022-06-01
Updated ES translation

RAVMN:


assets/_locales/es/messages.json
```diff
@@ -731,6 +731,10 @@
     "message": "Últimos enlaces de página usados",
     "description": "Last used page links"
   },
+  "settings_anilist_update_ui": {
+    "message": "Actualizar UI",
```
lolamtisch, line 735: The option adds a section to directly update an entry. I am not sure but something in the direction "UI de/para editar" is probably better here. Or what do you think?
image

lolamtisch (dismissed):

assets/_locales/es/messages.json
RAVMN: What do you think about "cuadro de actualización"(update box) or "bloque de actualización" (update block) ?
As it is not usual for final users to speak about "UI" in spanish

RAVMN: What do you think about "cuadro de actualización"(update box) or "bloque de actualización" (update block) ?
As it is not usual for final users to speak about "UI" in spanish

lolamtisch: `bloque de actualización` sounds fine to me. Is `cuadro` really used as a term for a UI box? If yes then I am fine with both.

henrik9999 (approved):

PR #1149, opened 2022-06-07
Addition of MangaBuddy

mac5617:


src/pages/MangaBuddy/main.ts
```diff
@@ -0,0 +1,85 @@
...
+    uiSelector(selector) {
+      j.$('div.summary')
+        .first()
+        .after(
+          j.html(
+            `<div id= "MALSyncheading" class="heading"> <span class="text-highlight">MAL-Sync</span></div><div id="malthing">${selector}</div>`,
+          ),
+        );
+    },
+    list: {
```
lolamtisch, line 60: The list is by default only shows part of the chapters on the overview page. The list extraction can be retriggered with `page.handleList();` function. Easierst would be to just listen on the show more button for a click and then get the new list  

https://github.com/MALSync/MALSync/blob/629efad20289c06df61ecf7f6efef367aba67ac0/src/pages/AnimeOnsen/main.ts#L129-L133

mac5617: Yeah, that fixed it. The series I initially tested had fewer chapters so I guess it didn't have any problems. How do you usually test pages before adding them?

Issue #1528, opened 2022-12-31
Doesn't update number of times rewatched


NorskNoobing:
### Describe the Bug

MAL-Sync won't update the number of times that you've rewatched an anime. Everything else seems to be working as intended, but the problem is that it just doesn't add 1 to the "Total Times Re-watched Series" field on the given anime.
![image](https://user-images.githubusercontent.com/39239702/210140030-80de52c6-ee6d-4872-91a4-e27a06b9927b.png)

See more details under the "Steps to reproduce the behavior" heading.

### Steps to reproduce the behavior

Start rewatching an already completed show by pressing "Ok" on the "Rewatch Anime" prompt.
![image](https://user-images.githubusercontent.com/39239702/210140107-61daf657-fec7-42b1-813f-cde4f98a1cf8.png)

You'll see that the "Re-watching" checkbox is checked on the anime, so that part works as intended.
![image](https://user-images.githubusercontent.com/39239702/210139936-d1c51467-2aa9-4283-aac6-d9ebc35ff280.png)

After finishing the re-watch press "Ok" on the "Finish rewatching" prompt.
![image](https://user-images.githubusercontent.com/39239702/210139866-281659f5-6bfd-47e3-ab4d-99f81402fc7e.png)

Then you'll see that the "Re-watching" checkbox's unchecked and the "Total Times Re-watched Series" field is still at the same number that it was originally set to (In my case that's "0").
![image](https://user-images.githubusercontent.com/39239702/210140052-2bec5011-3962-416d-bc9d-1fe424fd625a.png)
![image](https://user-images.githubusercontent.com/39239702/210140030-80de52c6-ee6d-4872-91a4-e27a06b9927b.png)

### Browser

Firefox

### Type

Extension

### URL

https://9anime.pl/watch/attack-on-titan-ova.4z4k/ep-3

PR #1533, opened 2023-01-03
fixed issue: Rewatch count not increased

1Hira0:
https://github.com/MALSync/MALSync/issues/1528


lolamtisch (changes requested):
Does not seem to work. The mal api thorws an error because of wrong parameter values
https://tenshi.moe/anime/igifz0rf/25
image
image

lolamtisch (approved):

PR #1540, opened 2023-01-06
Update AnimeOnsen

mist8kengas:
Update URLs + styles

### DIFF
--- a/src/pages/AnimeOnsen/meta.json
+++ b/src/pages/AnimeOnsen/meta.json
@@ -1,6 +1,6 @@
 {
-  "search": "https://animeonsen.xyz/search/{searchterm}",
+  "search": "https://www.animeonsen.xyz/search/{searchterm}",
   "urls": {
-    "match": ["*://animeonsen.xyz/*"]
+    "match": ["*://animeonsen.xyz/*", "*://www.animeonsen.xyz/*"]
   }
 }
--- a/src/pages/AnimeOnsen/style.less
+++ b/src/pages/AnimeOnsen/style.less
@@ -32,6 +32,10 @@ span.malp-group-label,
   border-color: currentcolor;
 }
 
+main div.container {
+  grid-template-rows: max-content repeat(2, auto) 1fr;
+}
+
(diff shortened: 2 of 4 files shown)


src/pages/AnimeOnsen/meta.json
```diff
@@ -1,6 +1,6 @@
 {
-  "search": "https://animeonsen.xyz/search/{searchterm}",
+  "search": "https://www.animeonsen.xyz/search/{searchterm}",
   "urls": {
-    "match": ["*://animeonsen.xyz/*"]
+    "match": ["*://animeonsen.xyz/*", "*://www.animeonsen.xyz/*"]
```
lolamtisch, line 4: `"*://*.animeonsen.xyz/*"` should match both

src/pages/AnimeOnsen/style.less
```diff
@@ -32,6 +32,10 @@ span.malp-group-label,
   border-color: currentcolor;
 }
 
+main div.container {
+  grid-template-rows: max-content repeat(2, auto) 1fr;
```
lolamtisch, line 36: What is the idea behind this styling?

lolamtisch (changes requested):

src/pages/AnimeOnsen/style.less
mist8kengas: > What is the idea behind this styling?

The page styling gets broken if the `malp` group label thingy gets inserted into the page. Because the details page has a specific `grid-template-rows` sizing, if a new element gets inserted in the grid, it will have malformed styling (in this case, `malp` will get `1fr` grid row size instead of `auto` and as such the elements of the description page (such as episode list) has no row size and is defaulted to minimum row size.

src/pages/AnimeOnsen/meta.json
mist8kengas: > `"*://*.animeonsen.xyz/*"` should match both

The site is strictly moving to using the `www.` subdomain, other subdomains are reserved for other functions that may not support MALSync. Weird things might happen if using the wildcard.

PR #1570, opened 2023-01-21
Well just updated readme.md and removed animixplay.to

LiReNa00:
Well since we all know that animix is now officially closed so I thought of updating the readme.md and pages.md 😓


lolamtisch:
This is not how it works and it is already done here.
https://github.com/MALSync/MALSync/pull/1517

LiReNa00:
Actually.... I never meant to change the code? Ofc ik that it has been removed in the code ... What i updated is readme cuz it still have animixplay name in it?

henrik9999:
> Actually.... I never meant to change the code? Ofc ik that it has been removed in the code ... What i updated is readme cuz it still have animixplay name in it?

The readme is autogenerated out of the code it doesn't have to be edited manually. It will get removed automatically from the readme when we do the nest release

PR #1590, opened 2023-01-30
Docchi.pl

anKordii:


src/pages/Docchi/main.ts
```diff
@@ -0,0 +1,99 @@
+import { pageInterface } from '../pageInterface';
+
+export const Docchi: pageInterface = {
+  name: 'Docchi',
+  domain: 'https://docchi.pl/',
```
henrik9999, line 5: Remove the / at the end

henrik9999 (changes requested):

src/pages/Docchi/main.ts
anKordii: fixed

anKordii:
@henrik9999 I'm first time doing something with prettier, color must be 1:1 `rgba(·0,·0,·7.5%)·`?

henrik9999:
> @henrik9999 I'm first time doing something with prettier, color must be 1:1 `rgba(·0,·0,·7.5%)·`?

Yes but the · are spaces, if you do `npm run lint:fix` it should probably fix it itself tho

henrik9999:
Btw dies the page have mal or anilist IDs?

anKordii:
> Btw does the page have mal or anilist IDs?

Yes, it has mal IDS

henrik9999:
> > Btw does the page have mal or anilist IDs?
> 
> Yes, it has mal IDS

Then you should implement getMalUrl will increase the correct title recognition

anKordii:
> 

Sure, I would do that from start if I know there option like that, but do you perhaps have a example code or could you tell me which website have it?

henrik9999:
https://github.com/MALSync/MALSync/blob/4e2e7599855650afc933e8dccd19a3bc4b911403/src/pages/YugenAnime/main.ts#L36

Something like that and it should on sync and overview

henrik9999:
Also if you don't like the selector stuff you could put everything into json like yugen does but that's up to you

anKordii:
> Also if you don't like the selector stuff you could put everything into a json like yugen does but that's up to you

For now, I will stick with selector but maybe in future I will switch to json

anKordii:
The timeout error probably is triggered because of too many requests page

anKordii:
> The timeout error probably is triggered because of special too many requests screen on website where no data is showed
> 
> image

You can redo checks, I changed some stuff on backend, so website shouldn't timeout

src/pages/Docchi/meta.json
```diff
@@ -0,0 +1,5 @@
+{
+  "urls": {
+    "match": ["*://docchi.pl/anime/*"]
```
lolamtisch, line 3: Because it is a single page app it needs to inject on all pages. `*://docchi.pl/*`

src/pages/Docchi/main.ts
```diff
@@ -0,0 +1,109 @@
...
+      return j.$('a[mal_sync="title"]').text();
+    },
+    getIdentifier(url) {
+      return Docchi.sync!.getIdentifier(url);
+    },
+    uiSelector(selector) {
+      j.$('div[mal_sync="episodes_list"]').before(j.html(selector));
+    },
+    getMalUrl(provider) {
+      return `https://myanimelist.net/anime/${j
```
lolamtisch, line 50: Does not work resulting in a error and nothing works. Please add a condition.

lolamtisch (changes requested):

src/pages/Docchi/meta.json
anKordii: Updated

src/pages/Docchi/main.ts
anKordii: I guess you was mentioning `getMalUrl`, I fixed it now returns ID perfectly

lolamtisch: Please an condition for animes that have no mal id.

lolamtisch (changes requested):

PR #1619, opened 2023-02-14
Fix: load myanimelist lists on safari/tampermonkey.

licebmi:
Tampermonkey on safari can't fetch the lists from MAL. Further debug indicates that GM_xmlhttpRequest can't parse the url, and url encoding the braces resolves the issue.

On my limited knowledge, I assume that url encoded strings are a safe bet; but I'm totally inexpert on web development and I'm not sure the effect on other browsers.


lolamtisch:
@licebmi Could you test if this pr fixes it? Because that would be a better solution.

#1656

licebmi:
Yep, that seems to be working just fine. Will close this PR then.

PR #1657, opened 2023-03-01
Add FMTeam

NatsuDzn:
Scantrad website is about to be closed, at the moment you can't even read anything on it
FMTeam is pretty much the successor of that website


viAtom (approved):
Looking good !

Care about tests, Batuque could get serialized and chapter 1 removed (happened to scantrad with dr. stone before), could be great to contact them for a testing series or anything that will last

PR #1735, opened 2023-04-05
Conversion to TypeScript

YggdrasillBR:
conversion to typescript all as possible


lolamtisch:
I have no idea how I should review this thing. Too many things, strange constructs, breaking other tools and often dont understand why the changes were made. I dont know what would be the best way here, probably to split every task into a seperate pr, but not sure

PR #1751, opened 2023-04-17
Update shikimori domain name

matryosha:
From .one to .me domain as one has been temp(?) lost

Source: https://shikimori.me/forum/news/505429-novyy-domen-sayta-shikimori-me Important part from there is:
  `В любом случае новый домен сайта теперь shikimori.me. Даже когда/если
  я верну старый домен, то основным доменом сайта останется новый, чтобы
  не устраивать еще один переезд.`
Which roughly translates as shikimori.me would the main domain even after restoring old ones.

### DIFF
--- a/src/_provider/Shikimori/helper.ts
+++ b/src/_provider/Shikimori/helper.ts
@@ -4,11 +4,11 @@ import { Cache } from '../../utils/Cache';
 
 const clientId = 'z3NJ84kK9iy5NU6SnhdCDB38rr4-jFIJ67bMIUDzdoo';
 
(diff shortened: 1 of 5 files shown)


matryosha:
@lolamtisch

lolamtisch:
We just had a release. Next release will be in a few weeks. I did not forget. 😉

matryosha:
From https://shikimori.me/forum/news/506738-novyy-domen-sayta-shikimori-me-v2

> [...]Обращения к апи [shikimori.one](https://shikimori.one/) будут работают без редиректа, чтобы не ломались приложения, использующие апи сайта. 

So the meaning that old domain redirects to new one so api access through .one works.

Either close it as it now works anyway or change the domain to main one in case something happens as it was last time.

lolamtisch:
Good to know. Yeah keep the PR open it makes sense to migrate to the new domain either way.

PR #1768, opened 2023-04-23
[Kaguya] Update domain

hoangvu12:
The website's domain was changed to kaguya.app.


henrik9999 (changes requested):
You need to edit it in the tests and meta as well

hoangvu12:
> You need to edit it in the tests and meta as well

I forgot that, sorry.

PR #1784, opened 2023-05-01
Add AnimeOnegai

IJCS:
Add suport for AnimeOnegai page.


src/pages/AnimeOnegai/main.ts
```diff
@@ -0,0 +1,140 @@
...
+    getTitle(url) {
+      return j.$('div.backButtonAssetInfo > div > small').text().trim();
+    },
+    getIdentifier(url) {
+      return id;
+    },
+    getOverviewUrl(url) {
+      return OVurl;
+    },
+    getEpisode(url) {
```
lolamtisch, line 37: Better to use regex group matching for this as it is easier and more reliable
Example : https://github.com/ijcs/malsync/blob/0d7efdf8bdc4bb35357149afc8e696f5dce48bde/src/pages/ADN/main.ts#L25

src/pages/AnimeOnegai/main.ts
```diff
@@ -0,0 +1,140 @@
...
+    j.$(document).ready(function () {
+      conseguir(page);
+    });
+    utils.urlChangeDetect(function () {
+      conseguir(page);
+    });
+  },
+};
+
+function conseguir(page) {
```
lolamtisch, line 93: it is bad practive to use variable/function names in other languages than english.

lolamtisch (changes requested):

src/pages/AnimeOnegai/main.ts
IJCS: I already changed it to an alternative like the one in the example... but gh did not detect it. Notice manually.

PR #1840, opened 2023-06-08
Add AnimeSub.LT page

PcmKomanda:


src/pages/AnimeSub/main.ts
```diff
@@ -0,0 +1,89 @@
...
+  },
+};
+
+function getId() {
+  const id = j.$('[rel="shortlink"]').attr('href')?.split('p=')[1];
+  if (id) return id;
+  return undefined;
+}
+
+function malId(provider) {
```
henrik9999, line 79: this function isnt used

henrik9999 (changes requested):

src/pages/AnimeSub/main.ts
PcmKomanda: Whoops. You're right. 😅

henrik9999: now you are not using the anilist link

henrik9999:
the site has anime and manga so you have to set the type in the init function accordingly `page.type = 'anime'` or `page.type = 'manga`

PcmKomanda:
> page.type = 'anime' or page.type = 'manga'

Property 'type' does not exist on type 'SyncPage'

lolamtisch:
Should be `AnimeSub.type` what can be set.
https://github.com/MALSync/MALSync/blob/master/src/pages/Proxer/main.ts#L119

PR #1841, opened 2023-06-08
Add 1stKissManga.me to page list

Hanawa02:


src/pages/list.json
henrik9999, line 1: This file is auto generated please don't commit

src/pages/FirstKiss/main.ts
```diff
@@ -0,0 +1,107 @@
...
+const URL_MANGA_ID_INDEX = 4;
+
+function isChapterPage(url: string): boolean {
+  return url.split('/')[5].startsWith('chapter');
+}
+
+export const FirstKiss: pageInterface = {
+  name: '1stKissManga',
+  domain: 'https://1stkissmanga.me',
+  languages: ['en-US'],
```
henrik9999, line 15: We are using the name so in this case it would be just "English"

henrik9999 (changes requested):

src/pages/list.json
Hanawa02: 💡 Thanks for pointing it out. I removed the changes

src/pages/FirstKiss/main.ts
Hanawa02: Updated

PR #1906, opened 2023-07-01
Remove websites taken down by DMCA.

uKaigo:
I only looked for Brazillian websites.
Note: [Muito Mangá](https://muitomanga.com) is not resolving, but I am not sure if it was taken down or if it's a temporary thing. But, if you want, I can remove it too.


uKaigo:
It was down (with a notice) when I first created the PR.
Also, should I include "Muito Mangá" in this pr?

PR #1938, opened 2023-07-12
feat: add Lucifer Donghua support

aadityataparia:
Adds support for anime (mainly chinese) website: https://www.luciferdonghua.co.in/ and https://www.luciferdonghua.in/


henrik9999 (changes requested):

src/pages/LuciferDonghua/main.ts
```diff
@@ -0,0 +1,86 @@
...
+    );
+
+    utils.waitUntilTrue(
+      () => {
+        return j.$('h1.entry-title');
+      },
+      function () {
+        con.info('Start check');
+        page.handlePage();
+        page.handleList();
```
lolamtisch, line 82: handleList is only necessary if the episode list is paginated. So it can be removed here.

src/pages/LuciferDonghua/main.ts
```diff
@@ -0,0 +1,86 @@
...
+    uiSelector(selector) {
+      j.$('.ts-breadcrumb').append(j.html(selector));
+    },
+    nextEpUrl(url) {
+      return j.$('a[rel=next]').attr('href');
+    },
+  },
+  overview: {
+    getTitle(url) {
+      return (j.$('h1.entry-title').text() || '').split('[')[0];
```
lolamtisch, line 41: Title always empty

src/pages/LuciferDonghua/main.ts
```diff
@@ -0,0 +1,86 @@
...
+  type: 'anime',
+  isSyncPage(url) {
+    return url.indexOf('-episode-') > -1;
+  },
+  isOverviewPage(url) {
+    return url.indexOf('/anime') > -1;
+  },
+  sync: {
+    getTitle(url) {
+      return (j.$('.det a').text() || '').split('[')[0];
```
lolamtisch, line 17: Does not work. Title is always empty.

src/pages/LuciferDonghua/main.ts
```diff
@@ -0,0 +1,86 @@
...
+        return Number(selector.children('.epl-num').first().text());
+      },
+    },
+  },
+  init(page) {
+    api.storage.addStyle(
+      require('!to-string-loader!css-loader!less-loader!./style.less').toString(),
+    );
+
+    utils.waitUntilTrue(
```
lolamtisch, line 75: I suspect that title etc does not work, because the page is not ready.
Try to wrap it into a `j.$(document).ready(function () {` to make sure the page is ready

lolamtisch (changes requested):

PR #1974, opened 2023-08-05
Add HinataSoul

uKaigo:
I couldn't run the tests, since it was taking way too long and it looked like it was frozen. But everything should work.

`paginationNext` wasn't running because `elementEp` was broken, but that should be fixed.

I'm testing for episode 37, but I'm not sure if pagination works in testing, will remove if it doesn't.

### DIFF
--- a/src/pages/HinataSoul/main.ts
+++ b/src/pages/HinataSoul/main.ts
@@ -1,0 +1,95 @@
+import { pageInterface } from '../pageInterface';
+
+export const HinataSoul: pageInterface = {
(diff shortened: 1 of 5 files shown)


uKaigo:
I don't know why page tests are failing. It works when running the extension.

lolamtisch:
Is normal

Source-Dom:
Btw your `pages.ts` is missing a semicolon `;` on your 135th line.

lolamtisch (changes requested):
Fix the missing ;

Issue #1971, opened 2023-08-05
MALSync fails to recognize oneshots while using komga.


garvongithub:
### Describe the Bug

Recently komga has introduced a new way to seperate series with multiple books/chapters and single unrelated books/chapter (oneshots). Before it was only series and that series would have one book but now it can display single books/chapter without having to put them in series. The Url has changed from http://localhost:25600/series/* to http://localhost:25600/oneshot/* for those single books/chapter so MALSync fails to recognize them.

### Steps to reproduce the behavior

1. Make a oneshot directory in Komga library folder and move archives there.
2. Enable oneshots and provide the location in library settings.
3. Open any book/chapter from oneshot directory.

### Browser

Firefox

### Type

Extension

### URL

http://localhost:25600/oneshot/*

PR #1984, opened 2023-08-14
Updated Komga

rRoler:
Added:
- a way to handle oneshots (#1971)
- detection for onsite mal, al and kt links
- readerConfig (only for webtoon reading mode, can't make it work on other ones)

Changed:
- use the series title from its metadata for searching on sync pages instead of the series name
- ignore incognito reading


lolamtisch:
@rRoler Everything was fine. I changed the reader config to support all ready types. Could you review my changes? After that everything should be ready

lolamtisch (approved):

Issue #977, opened 2021-11-23
[Website Request] <Website Name>


buqcay:
## Website
- Name: coloredmanga
- Link: [?](https://coloredmanga.com/)
- Language(s): English

buqcay:
when i first created this issue i forgot to put the name of the site in the title. the title is now updated but the github bot's message in the discord is still outdated/incorrect.

PR #1997, opened 2023-08-22
Added ColoredManga

fabiomurru96:
Added [ColoredManga](https://coloredmanga.com/) to supported pages, it derives from Isekaiscan as they're both implemented using [madara](https://github.com/guendev/madara).

I've slightly refactored Isekaiscan's page to be able to clone and override it (took inspiration from [MangaHere](https://github.com/MALSync/MALSync/blob/master/src/pages/MangaHere/main.ts))

---

Related Issue
- https://github.com/MALSync/MALSync/issues/977

### DIFF
--- a/src/pages/ColoredManga/main.ts
+++ b/src/pages/ColoredManga/main.ts
@@ -5,19 +5,17 @@ const clone = getInter();
 
 clone.name = 'ColoredManga';
 clone.domain = 'https://coloredmanga.com';
-clone.isSyncPage = function(url) {
-    return Boolean(
-      utils.urlPart(url, 3) === 'mangas' &&
-        (utils.urlPart(url, 5) &&
-        utils.urlPart(url, 5).startsWith('chapter-') ||
-        utils.urlPart(url, 6) &&
-        utils.urlPart(url, 6).startsWith('chapter-'))
-    );
-}
+clone.isSyncPage = function (url) {
+  return Boolean(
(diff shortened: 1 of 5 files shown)


fabiomurru96:
The Isekaiscan diff is pretty bad here in github, if it can help reviewing these are the real changes 

<details>

<summary>Click here to see image</summary>

image
</details>

src/pages/ColoredManga/main.ts
```diff
@@ -0,0 +1,65 @@
+import { pageInterface } from '../pageInterface';
```
lolamtisch, line 1: Not all Chapters are recognized

https://coloredmanga.com/mangas/medaka-box/volume-12/box-106/
https://coloredmanga.com/mangas/spy-x-family-colour/volume-10-bw/mission-67-2-bw/
https://coloredmanga.com/mangas/eng-rurouni-kenshin/volume-28/act-255-finale/

src/pages/ColoredManga/main.ts
```diff
@@ -0,0 +1,65 @@
...
+import { getInter } from '../Isekaiscan/main';
+
+const clone = getInter();
```
lolamtisch, line 4: Title could be cleaned up a bit to make the matching with mal better.
Full Color
{Official TL}
(EN)

lolamtisch (changes requested):

PR #2006, opened 2023-08-25
Add Honeymanga [UA] website

aTei:
Ukrainian manga website support.

Sadly, it doesn't have search page and only json endpoint

https://honey-manga.com.ua/

image

### DIFF
--- a/src/pages/HoneyManga/main.ts
+++ b/src/pages/HoneyManga/main.ts
@@ -1,0 +1,87 @@
+import { pageInterface } from '../pageInterface';
+
+export const HoneyManga: pageInterface = {
+  name: 'HoneyManga',
+  domain: 'https://honey-manga.com.ua/',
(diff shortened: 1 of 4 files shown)


aTei:
Is it possible to run tests only for HoneyManga page? When I try to test it collapses on other pages before it gets to this one.

NoPlagiarism:
Can you add HoneyManga to [pages.ts](https://github.com/MALSync/MALSync/blob/master/src/pages/pages.ts) and commit it?

> Is it possible to run tests only for HoneyManga page?

Change [this line](https://github.com/MALSync/MALSync/blob/3615530897cf96ef31a431bf02aaa41e716ca852/test/headless/test.js#L258)

Also, you can join MAL-Sync DIscord server: https://discord.com/invite/cTH4yaw

PR #2016, opened 2023-08-31
Sticky Notifications option

MulverineX:


src/background/upgradeWizzards.ts
MulverineX, line 1: Should just need this set, not sure what the next version will be.

src/_minimal/components/settings/settings-structure-estimation.ts
```diff
@@ -35,6 +35,19 @@ export const notificationsSection: ConfObj[] = [
     },
     component: SettingsGeneral,
   },
+  {
+    key: 'notificationsSticky',
+    title: () =>
+      `${api.storage.lang('settings_Notifications_Sticky')} ${api.storage.lang(
+        'settings_Notifications',
+      )}`,
+    condition: () => true,
```
MulverineX, line 44: Not sure if `condition` is optional in this case, or whether I should make it under the same condition as the other settings.

src/_minimal/components/settings/settings-structure-estimation.ts
```diff
@@ -35,6 +35,19 @@ export const notificationsSection: ConfObj[] = [
     },
     component: SettingsGeneral,
   },
+  {
+    key: 'notificationsSticky',
+    title: () =>
+      `${api.storage.lang('settings_Notifications_Sticky')} ${api.storage.lang(
```
lolamtisch, line 41: This will not work in all languages. Better use "sticky notifications" as translation string

src/_minimal/components/settings/settings-structure-estimation.ts
lolamtisch, line 44: It should be optional

lolamtisch (changes requested):

src/_minimal/components/settings/settings-structure-estimation.ts
MulverineX: On what?

lolamtisch: You can just remove the condition completly

PR #2019, opened 2023-09-02
AnimetTV source added

asadbek064:
New source added AnimetTV
Gogo 
aniwatch
4K Anime Title
etc


lolamtisch (changes requested):
You forgot to add it to src\pages\pages.ts

PR #2124, opened 2023-11-04
[AnimeGO] fix getTitle

NoPlagiarism:
There are few anime overviews that doesn't have English and Japanese title

Example: https://animego.org/anime/oshi-no-ko-2307 \(Obviously, [Oshi no Ko](https://anilist.co/anime/150672/Oshi-No-Ko/)\)

### Before fix:
- On 0.49% overviews extension crashes (no alternative headlines at all or only English)
- 30.56%: Japanese (日本語) headline is chosen

### After fix:
- Uses Russian title if no title found
- Locating English title correctly


NoPlagiarism:
I'm gonna check all of site DataBase to make sure that every title will be chosen correctly

NoPlagiarism:
Now better

PR #2215, opened 2023-12-28
Add Taiyo Page

Luscarvalho:


src/pages/pages.ts
```diff
@@ -1,3 +1,4 @@
+import { Taiyo } from './Taiyo/main';
```
henrik9999, line 1: move this to the bottom

src/pages/Taiyo/main.ts
```diff
@@ -0,0 +1,120 @@
...
+      return true;
+    }
+    return false;
+  },
+  sync: {
+    getTitle(url) {
+      return j.$('.media-title').text();
+    },
+    getIdentifier(url) {
+      return url.split('/')[4];
```
henrik9999, line 35: the identifier has to be the same as on the overview

henrik9999: you could do return Taiyo.overview!.getIdentifier(Taiyo.sync.getOverviewUrl(url));

PR #2275, opened 2024-02-15
New comick domain

Uranite:
Please point out if there is mistakes


kusuoh:
comick change domain again.

new domain: https://comick.io/

Uranite:
> comick change domain again.
> 
> new domain: https://comick.io/

Yes, that is the point of this pr

PR #2286, opened 2024-02-21
Add Suwayomi

lolamtisch:
Extends #2277


lolamtisch:
@TriForMine I added some fixes. Would be nice if you could test them

voltrare:
seems good tracking working for me

PR #2342, opened 2024-03-17
Update AniWatch to HiAnime

vvirtues:
aniwatch -> hianime


henrik9999:
This file is auto generated and will be updated with the next release

PR #2382, opened 2024-04-21
Fix: Guya/Cubari chapter progress detection

xaskii:
Fixes chapter progress detection on guya/cubari.moe when using RTL (right-to-left reading mode). The reader reverses the order of the page elements, so `countAbove` on the `.ReaderImageWrapper` selector reports the wrong page number. The selector worked correctly when reading vertically and left-to-right.

### DIFF
--- a/src/pages/Guya/main.ts
+++ b/src/pages/Guya/main.ts
@@ -46,8 +46,9 @@ export const Guya: pageInterface = {
     readerConfig: [
       {
         current: {
-          selector: '.ReaderImageWrapper',
-          mode: 'countAbove',
+          selector: '.rdr-page-selector-counter',
+          mode: 'text',
+          regex: '^\\d+',
         },
         total: {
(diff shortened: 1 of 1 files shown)


src/pages/Guya/main.ts
```diff
@@ -46,8 +46,9 @@ export const Guya: pageInterface = {
     readerConfig: [
       {
         current: {
-          selector: '.ReaderImageWrapper',
-          mode: 'countAbove',
+          selector: '.rdr-page-selector-counter',
+          mode: 'text',
+          regex: '^\\d+',
         },
         total: {
```
lolamtisch, line 53: Would it not be better in this case to use the .rdr-page-selector-counter for the total too. So that it uses the same place for both?

lolamtisch (changes requested):

henrik9999 (approved):

PR #2408, opened 2024-05-16
fix new style monoschinos

MixDevCode:


MixDevCode:
Fixed in https://github.com/MALSync/MALSync/pull/2416

PR #2464, opened 2024-06-13
Fix episode detection not working after Miruro player page url changes

Debjoy:
Hi,
Episode detection had stopped working in Malsync after Miruro.tv made changes to their URL for video players.
eg. 
from: https://www.miruro.tv/watch/101280/tensei-shitara-slime-datta-ken/20 
to: https://www.miruro.tv/watch/101280/20

the episode numbers are taken from url itself, so that logic had to be changed.

Best Regards,
Debjoy

### DIFF
--- a/src/pages/Miruro/main.ts
+++ b/src/pages/Miruro/main.ts
@@ -6,7 +6,11 @@ export const Miruro: pageInterface = {
   languages: ['English'],
   type: 'anime',
   isSyncPage(url) {
-    return utils.urlPart(url, 3) === 'watch' && utils.urlPart(url, 5).trim() !== '';
+    return (
+      utils.urlPart(url, 3) === 'watch' &&
+      utils.urlParam(url, 'id') !== null &&
+      utils.urlParam(url, 'ep') !== null
+    );
   },
   isOverviewPage(url) {
     return false;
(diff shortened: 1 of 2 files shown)


Debjoy:
Making this PR a draft for now.. as Miruro.tv is still ongoing changes related to watch page. Will update once the changes in Miruro is final.

Debjoy:
Hi, 
I have updated the PR with the modified changes. Now the url shceme is `https://www.miruro.tv/watch?id=139630&ep=1` . I have confirmed that this is final.
I also did one modification to the waitUntilTrue to check if it has "ep" query param in it, as when clicked on an anime, it changes the url of the page from  `https://www.miruro.tv/watch?id=139630` to  `https://www.miruro.tv/watch?id=139630&ep=1` which triggers the urlChangeDetect() again. This was done to avoid calling of handlePage() two times for the same page.

Debjoy:
Apparently we have an overview page as well.. I need to add it. will create a commit soon. making the PR draft for now

Debjoy:
Added support for over view page.

src/pages/Miruro/main.ts
```diff
@@ -57,14 +61,30 @@ export const Miruro: pageInterface = {
...
       require('!to-string-loader!css-loader!less-loader!./style.less').toString(),
     );
     let inte: NodeJS.Timer;
     utils.urlChangeDetect(() => ready());
-    j.$(document).ready(() => ready());
+    j.$(() => ready());
     function ready() {
+      if (!Miruro.isSyncPage(window.location.href) && !Miruro.isOverviewPage!(window.location.href))
+        return;
       page.reset();
```
lolamtisch, line 88: Please move the page.reset in front of the condition. This is necessary to remove malsync UI when opening an page that is not the sync or overview page.

lolamtisch (changes requested):

src/pages/Miruro/main.ts
Debjoy: I have done the changes

PR #2527, opened 2024-07-26
Cleanup VOE no longer valid domains

louisroyer:
Remove every VOE domain (except voe.sx) no longer responding with HTTP 302 redirection to `https://voe.sx/`. Either because:
- they no longer resolve
- they resolve but no server is listening on port 443
- they are responding HTTP 302 redirection to domain unrelated to VOE
- they are responding HTTP 200 with content unrelated to VOE

It would be better to do this automatically by the way :wink:


louisroyer:
Ps: I don't know if it is possible to remove permissions for these domains automatically.

PR #2546, opened 2024-08-10
Fix malsync not working for Miruro in overview page

Debjoy:
Hi, due to recent changes in the site Miruro, malsync had stopped working in overview page. 
I did few modification to change mostly the selectors and some conditions to make it work again.
There is a minor css fix as well for the malsync ui.

Best,
Debjoy


Debjoy:
Hi the test is failing because of the pages are cached.. Can anything be done in that ase?

Issue #1753, opened 2023-04-17
[Website Request] Scylla Scans


crimsonkatana99:
### Website name

Scylla Scans

### Website link

https://scyllascans.org/releases

### Website language(s)

English

### Other details

High-quality English-language manga / manhua / manhwa scan group main website.

Pranav-uchiha:
Stop

PR #2586, opened 2024-09-03
Support for ScyllaScans and VortexScans

SoldierZoom:
Adds support for tracking on https://scyllacomics.xyz/ and ~https://mangagalaxy.net/~ https://vortexscans.org/


SoldierZoom:
Also this should close #1753

PR #2588, opened 2024-09-06
Add AniwatchTV support

RaSan147:


lolamtisch:
We do not add fake/clone websites. Please use the official page https://hianime.to/

RaSan147:
It was down 💀
So got scared and added it.

Issue #2061, opened 2023-09-25
[Website Request] LibGroup (Manga) [RU]


NoPlagiarism:
### Website name

[MangaLib](https://mangalib.me), [SlashLib](https://slashlib.me)

### Website link

https://mangalib.me, https://slashlib.me

### Website language(s)

Russian

### Other details

It's big chain of websites that contains Russian translation for Manga. For simplicity "LibGroup" (Manga**Lib** and etc)

- One profile for all of websites, so it should be the same back-end for both of domains
- It has an API. 
    - For example. for ONIMAI ([AniList](https://anilist.co/manga/100080/ONIMAI-Im-Now-Your-Sister/), [MangaLib](https://mangalib.me/onii-chan-wa-oshimai)) it will be: `https://mangalib.me/manga-short-info?slug=onii-chan-wa-oshimai`
    - [Unofficial API Docs](https://github.com/NekonexDev/MangaLibAPI), [MangaLibAPI on Python](https://github.com/ArhostCode/MangaLibApi)

NoPlagiarism:
Each overview and read page has `<script>` with `window.__DATA__ = ` with a lot of data.

Also, reading page has `window.__info = ` with data about current chapter, volume

NoPlagiarism:
Temporarily closed until UI update.

For reference links to new UI:
[MangaLib](https://test-front.mangalib.me), [YaoiLib \(SlashLib\)](https://test-front.slashlib.me), [HentaiLib](https://test-front.hentailib.me)

NoPlagiarism:
Reopening in case if anyone wants to take old UI or/and to escape any duplicates

DFreezerR:
Do you think ranobe one should be implemented? I don't know if malsync have even a single light novel website support. But I think it might be possible since providers have light novels tracking. 

It just happened that I made some work on AnimeLib using their API, and since all of their websites uses the same UI + API, it will be much easier to implement others. Do you, by chance, have some work in progress regarding these websites? (Although I should've ask that before my PR...)

henrik9999:
Btw Malsync already fully supports light novels, we just dont have any LN site currently implemented

DFreezerR:
> Btw Malsync already fully supports light novels, we just dont have any LN site currently implemented

Good to know!

NoPlagiarism:
> Do you, by chance, have some work in progress

Nope, I have no skills enough and too much perfectionism in my blood to deal with this SPA

[...]

PR #2606, opened 2024-09-19
Animelib sync

DFreezerR:
Added sync to AnimeLib. Website upgraded it's UI and I decided to add it.  PR related to #2061 because it's a part of these websites. Had to add new player to `src/pages/playerUrls.js`  since it's their own player.
If there's someone who want to do it more profound way, say it and feel free to close it (telling this since original issue mentioned APIs and other websites in that group)


DFreezerR:
need to fix selectors to api where I can since html classes are changing (not familiar with vue, I guess it changes every build?)

DFreezerR:
Works now

lolamtisch:
@DFreezerR 
Calling apis is always a bit of gamble, because of CORS etc.

In hidive we experimented with overriding the fetch of the website in proxy script and listening so on the requests the website does. No complain yet from the users. Do you think that could be an option too here?

https://github.com/MALSync/MALSync/blob/master/src/pages/Hidive/proxy.ts

DFreezerR:
> @DFreezerR Calling apis is always a bit of gamble, because of CORS etc.
> 
> In hidive we experimented with overriding the fetch of the website in proxy script and listening so on the requests the website does. No complain yet from the users. Do you think that could be an option too here?
> 
> https://github.com/MALSync/MALSync/blob/master/src/pages/Hidive/proxy.ts

The main issue here (AnimeLib, MangaLib, RanobeLib) was that users can not access some title's chapters due to 2 options:
1. Title might be licensed so chapters/episodes got deleted - you can do nothing here
2. Title might be only read/watched if you auth into account - create account

Currently I made a logic getting title's info based on API, but if it fails, the code will try get info using selectors. So like a backup method in case user trying to read\watch something required an auth.

I think it's possible to do what you mean here, I just didn't see this option. Guess might add it like a 3rd variant just in case. Do you think overriding fetch would be better approach? I manly did API because several websites are using it + it was convenient.

lolamtisch:
Then let's keep it like it is until it breaks with api

DFreezerR:
> Then let's keep it like it is until it breaks with the api

Sure.
Even if API breaks, the selectors will back it up for some time.
The worst case scenario if API breaks:
1. AnimeLib - Can not get season number on sync page - optional
2. MangaLib - Can not get cover image on sync page - still optional / Can not get total number of chapters - also optional 
3. RanobeLib - same as p.2

Ideally yes, there should be something like you mentioned in case "some situations"
I'll look into this thing you linked and try implement it as "additional improvements", but a bit later

src/pages/AnimeLib/main.ts
```diff
@@ -0,0 +1,261 @@
...
+            return j.$('.tabs-item').length;
+          },
+          () => {
+            page.handlePage();
+          },
+          500,
+        );
+      }
+
+      page.handlePage();
```
lolamtisch, line 130: This results in all overivew pages beeing triggered 2 times. I imagine you wanted inside of  https://github.com/MALSync/MALSync/pull/2606/files#diff-ad21d95b5e6745de2f7f26e52df7f79efe09ffce69150b485a72cb64e069ca45R102

src/pages/playerUrls.js
```diff
@@ -741,4 +741,8 @@ module.exports = {
   doflix: {
     match: ['*://doflix.net/*'],
   },
+  // anilib
+  anilib: {
```
lolamtisch, line 745: I did not find any player that would need this. Do you have an example. If the player is directly on the page without an iframe then no additional injection is necesseary.

src/pages/AnimeLib/main.ts
DFreezerR: Ah, yes, this need to be in sync one

src/pages/playerUrls.js
DFreezerR: Was long time since then, was thought that any new player should be added manually, fixed it

lolamtisch (approved):

Issue #2568, opened 2024-08-27
Remove Dead Websites


EmeraldStar0:
### Suggestion

Websites include Anix, AnimeSuge, and Aniwave, delay this issue however until we know they will be permanently down.

PR #2618, opened 2024-09-28
Aniwave/Anix/AnimeSuge Removed from MAL-Sync

mega6382:
Solves #2568


SimpliAj:
thre is now aniwave.se

henrik9999:
> there is now aniwave.se

Fake site, will not be added

PR #2624, opened 2024-10-01
Mangalib sync

DFreezerR:
Added MangaLib sync. Part of the #2061 issue and can count like [2/5] websites. 

Added both domains for manga, and two domain for SlashLib (non-hentai yaoi manga)

PR contains `src/pages/AnimeLib/api.ts` file from another PR, but I also added it here just in case. This is like a dependency for this website, did this because I don't think duplicate files in several directories are necessary. Until the file is merged I could keep them the same in both PRs

Could not implement chapter list handler since page uses **vue-virtual-scroller** to handle the list. It basically loads new objects when user scrolls and unloads older ones at the same time, turning the task to get all objects from the list kinda unreal. 

<h3>Like the other PR (#2606) this one takes access tokens from authorized user to make API calls. I guess there "might" be security problems if someone tries something bad this way. And if this breaches this projects ethical or security conducts, please tell. Just keeping everyone informed just in case.</h3>


DFreezerR:
In Last commit([e9394a3](https://github.com/MALSync/MALSync/pull/2624/commits/e9394a39c8267d4901942c13f482df3c8e94bcbb)) I added ability to count chapters with float numbers for manga progress. To simplify, here's some cases:

1.  Normal chapters   - 1, 2, 3, 4, 5 
Will be counted normally by `readerConfig` where `current` and `total` number of pages come from html elements
2. Subchapters           -  _1.1, 1.2, 1.3, 1.4, 1.5_
These are still counter as chapter 1, but it's divided by translators to separate uploads. Each of them will be counted as "page" rather than chapter. So the current chapter will be counted as finished only if you read the last (in this case 1.5) "chapter". If user have "**Sync manga after 90% of the pages**" disabled, the chapter will be counted as completed on the first page (in this case 1.1). Also I bypassed 90% limit from `MangaProgress` class cuz in cases where we have chapters 1.1 and 1.2, the chapter will be completed instantly on "chapter" 1.1.
3. "Bonus" chapters    - 1, 2, 3, **3.5**, 4, 5
If there are chapters with float point number while there's still "normal" chapter (in this case 3), they will not be counted and progress will go as normal page by page
4. Mixed                      - 1, 2, _2.1, 2.2, 2.3_, 3, 4, 5, **5.5**
All above rules will be working. 1-2, 5 chapters progress counts as normal, 2.1-2.3 counts as "whole", and 5.5 won't be count at all

Some websites might upload chapters from different translators in the same pile, for example - 10, 11, 12.1, 12.2 where 12.1 and 12.2 are the SAME chapter but with different translator/language/fixes/etc. This situation is not present on MangaLib as far as I can see.

If this "feature" will be breaking I'll disable it

DFreezerR:
Since all new design websites from #2061 will have same UI, this PR will "close" 3/5 websites (MangaLib, SlashLib, HentaiLib (users can potentially add it with custom domains)). With that, merging this PR + #2631 + #2606 will close original issue.

src/pages/MangaLib/main.ts
```diff
@@ -0,0 +1,300 @@
...
+    getEpisode(url) {
+      return manga.reader.chapter;
+    },
+    getVolume(url) {
+      return manga.reader.volume || 1;
+    },
+    nextEpUrl(url) {
+      return manga.reader.next;
+    },
+    readerConfig: [
```
lolamtisch, line 70: I feel like this is a bit overcomplicated. I think it would be better to just disable updates generally if not at the last subchapter and use the normal page based config on the last subchapter. What do you think about that?


Currently the condition is only a dom selector. But it could easily be extended to an boolean function. 

```
readerConfig: [
      {
        condition: () => {
          if (not float) return false;
          if (not able to get total_subchapters or 0 or whatever is important) return false;
          return manga.reader.total_subchapters > manga.reader.current_subchapter_index! + 1;
        },
        {
        current: {
          mode: 'callback',
          callback: () => 0,
        },
        total: {
          mode: 'callback',
          callback: () => 2,
        },
      },
      {
        current: {
          selector: 'footer',
          mode: 'text',
          regex: '(\\d+) / (\\d+)$',
          group: 1,
        },
        total: {
          selector: 'footer',
          mode: 'text',
          regex: '(\\d+) / (\\d+)$',
          group: 2,
        },
      },
    ],
```

src/pages/MangaLib/tests.json
```diff
@@ -0,0 +1,31 @@
+{
+  "title": "mangalib",
+  "url": "https://mangalib.org/",
+  "testCases": [
+    {
+      "url": "https://mangalib.org/ru/3757--ijiranaide-nagatoro-san/read/v18/c132",
```
lolamtisch, line 6: I think it was hit by a DCMA takedown

src/pages/MangaLib/main.ts
```diff
@@ -0,0 +1,300 @@
+/* eslint-disable new-cap */
+/* eslint-disable global-require */
+import { pageInterface } from '../pageInterface';
+import { SyncPage } from '../syncPage';
+import { Manga, getMangaData, getChaptersData, isPageAPI, Chapters } from '../AnimeLib/api';
```
lolamtisch, line 5: I merged it already. But is the bearer token really necessary? I did not see anything yet that would need it. It is a bit scary that the api is a different tld. Do you think it would be possible without it?

https://github.com/MALSync/MALSync/blob/master/src/pages/AnimeLib/api.ts#L257

lolamtisch (changes requested):

src/pages/MangaLib/main.ts
DFreezerR: I guess what you are talking is more simpler, right. I'll try it and check how it feels, I might change it to what you mean in both PRs since I did it the same way

src/pages/MangaLib/tests.json
DFreezerR: Well.. you are right. Will change it. Local company got license to translate/sell manga so they removed it

src/pages/MangaLib/main.ts
DFreezerR: I included it just in case since I didn't know which PR you will merge first.

> But is the bearer token really necessary? I did not see anything yet that would need it

There are some manga which require you to auth to read it. There are not a lot of it but taking the token it the only thing I could think of beside implementing oauth. There's an example of said manga - https://mangalib.me/ru/manga/8359--watermelon-
Without auth you cant read it + and only API data you can get is manga info (so no chapters data). That means if user will try to read it, the whole "subchapter system" will not work, so to ensure 100% stability I did this token thing. I can remove it if you think this is too dangerous. 

> It is a bit scary that the api is a different tld

What so scary about it? You mean it could be some 3rd party api? They use this api on all they websites for manga/anime/novels. Without using API there's no way to get cover image + chapter info + chapters count on sync page

src/pages/MangaLib/main.ts
DFreezerR: Btw how did you write condition as function? It's only accepts `string`

lolamtisch (approved):

PR #2625, opened 2024-10-02
Kickass anime sync fixed

mega6382:
Fixes the KickAssAnime sync issue based on the new site structure.


lolamtisch (changes requested):
Extension can not directly access the dom variables you will need to implement an proxy script for that. But I do not see anything broken with the currect implementation?

mega6382:
The structure for the site changed in the beginning of October when I created this PR. But it is back to normal now. I didn't observe it during the past month, so I am not sure when that happened. But this is not needed anymore.

PR #2631, opened 2024-10-07
Ranobelib sync

DFreezerR:
Part of #2061 [3/5]
RanobeLib - LN + WN website.
Did almost everything based on #2624 
Includes api.ts file from #2606 
Had to modify internal files a bit to allow search + correction search work with SHIKI sync mode and novel types. Because as it is, there's no way to search novels on SHIKI sync mode. To do it, shikimori have a separate page `/ranobes` while other providers use `/manga` page to track both manga and novels. To mess as less as I could with internals I used `options` parameter in `searchInterface` and passed it down from  `src/_provider/Search/vueSearchClass.ts` with the novel type check to determine if the page in novel type. I "could've" add another page type in `pageInterface` but that would've meant I will need to change types almost everywhere, and to reduce changes I used already existed `options` and `this.type` in  `src/_provider/Search/searchClass.ts` and passed it down to search function. Moreover the only provider that uses this check is shikimori so there's no need to change every search function.
Additionally changed how `src/_provider/Search/searchClass.ts` works a bit since to search for novel more precisely you need to include both 'novels' and 'light novels' + added filter to search only for either of them.

As I tested, there are not much novels on either providers, and you WILL encounter situation where you either have to specify novel by correction search or to use local sync. Or you can add your novel on MAL yourself (or on any other provider). To me, the probability of novel existing was around 60%, so - <h4>If you gonna use it, always check if novel was set correctly in either MAL-Sync uiSelector on overview page or using float button in bottom right</h4>

Also,
<h3>Like the other PRs(#2606 + #2624) this one takes access tokens from authorized user to make API calls. I guess there "might" be security problems if someone tries something bad this way. And if this breaches this projects ethical or security conducts, please tell. Just keeping everyone informed just in case.</h3>


DFreezerR:
uh...

src/_provider/Search/components/search.vue
```diff
@@ -82,6 +82,14 @@ export default {
       type: Number,
       default: 0,
     },
+    options: {
```
lolamtisch, line 85: Remove as the filtering by novel happens afterwards on the "isNovel". This allows for the automatching to only match novels. The user facing search Ui does not have the filtering, making it possible for the user to select an novel for manga pages if they have one for some reason.

src/_provider/Search/components/search.vue
```diff
@@ -119,7 +127,7 @@ export default {
       if (this.searchKeyword) {
         this.loading = true;
 
-        normalSearch(this.searchKeyword, this.type).then(items => {
+        normalSearch(this.searchKeyword, this.type, this.options).then(items => {
```
lolamtisch, line 130: revert

src/_provider/Search/correctionApp.vue
```diff
@@ -56,6 +56,7 @@
         :type="searchClass.getNormalizedType()"
         :sync-mode="Boolean(syncMode)"
         :current-id="searchClass.getId()"
+        :options="options"
```
lolamtisch, line 59: revert

src/_provider/Search/correctionApp.vue
```diff
@@ -81,6 +82,7 @@ export default {
     inputOffset: 0 as number | '0',
     minimized: false,
     syncMode: null,
+    options: {},
```
lolamtisch, line 85: revert

src/_provider/Search/vueSearchClass.ts
```diff
@@ -39,6 +39,9 @@ export class SearchClass extends SearchClassExtend {
       });
       this.vueInstance.searchClass = this;
       this.vueInstance.syncMode = syncMode;
+      this.vueInstance.options = {
```
lolamtisch, line 42: revert

src/_provider/Shikimori/search.ts
```diff
@@ -4,11 +4,13 @@ import * as helper from './helper';
 export const search: searchInterface = async function (
```
lolamtisch, line 4: This is too much of a workaround. Better to revert this.

As far as I see the real fix would be to migrate Shikimori to the graphql endpoint, which does not have the novel destinction anymore. But that would be better in a seperate PR

src/utils/Search.ts
```diff
@@ -4,8 +4,9 @@ import { searchResult } from '../_provider/definitions';
 export async function normalSearch(
   searchterm: string,
   type: 'anime' | 'manga',
+  options = {},
```
lolamtisch, line 7: revert

lolamtisch (changes requested):
Sorry for the late review. Was too bussy and changes to the search always need thorough review.

src/_provider/Search/searchClass.ts
```diff
@@ -370,10 +371,12 @@ export class SearchClass {
     let url = `https://myanimelist.net/${this.getNormalizedType()}.php?q=${encodeURI(
       this.sanitizedTitel,
     )}`;
+    // NOTE - Novels on MAL can be:
+    // Light novel - type=2
+    // Novel       - type=8
+    // So best to search all types and then filter them
     if (this.type === 'novel') {
```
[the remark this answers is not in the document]
DFreezerR: True, you are right

src/_provider/Search/searchClass.ts
```diff
@@ -44,6 +44,7 @@ export class SearchClass {
     protected identifier: string,
   ) {
     this.identifier += '';
+    // REVIEW - Proposal, can we rename every 'titel' to 'title' except DE locale?
```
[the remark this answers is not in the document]
DFreezerR: Done

src/_provider/Search/components/search.vue
DFreezerR: Reverted whole shikimori changes

src/_provider/Search/components/search.vue
DFreezerR: Reverted whole shikimori changes

src/_provider/Search/correctionApp.vue
DFreezerR: Reverted whole shikimori changes

src/_provider/Search/correctionApp.vue
DFreezerR: Reverted whole shikimori changes

src/_provider/Search/vueSearchClass.ts
DFreezerR: Reverted whole shikimori changes

src/utils/Search.ts
DFreezerR: Reverted whole shikimori changes

src/_provider/Search/searchClass.ts
```diff
@@ -392,6 +395,15 @@ export class SearchClass {
               return handleResult(response, i + 1, This);
             }
           }
+          if (This.type === 'novel') {
```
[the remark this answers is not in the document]
DFreezerR: I can think of something like this:
```ts
if (This.type === 'manga' || This.type === 'novel') {
  const typeCheck = response.responseText.split(`href="${link}" id="si`)[1].split('</tr>')[0];
  const linkIsNovel = typeCheck.indexOf('Novel') !== -1;

  if ((This.type === 'manga' && linkIsNovel) || (This.type === 'novel' && !linkIsNovel)) {
    logger.log(`${linkIsNovel ? 'Novel Found' : 'Novel Not found'} check next entry`);
    return handleResult(response, i + 1, This);
  }
}
```
?

src/_provider/Shikimori/search.ts
DFreezerR: Btw I have a test implementation of graphql, can send a PR so you can check if this is worth working with

lolamtisch: I did not check it out in detail, but if it uses the same Auth token I do not see an issue with mixing it. Specially because I imagine it is easy to split. The user lists and search can use graph and the single can use the rest API, as it handles novel IDs at the manga endpoint too.

DFreezerR: got it

lolamtisch (approved):

PR #2640, opened 2024-10-12
Fix fullscreen button in MAL in anime page

Source-Dom:
In MyAnimeList when you go on the page of an anime, the videos/trailers/op's/etc. don't have the `allow="fullscreen"` attribute  in the iframe, which makes it so the fullscreen button of the embed youtube video is disabled. 

This enables it by adding `allow="fullscreen"` before the `src` attribute which enables the fullscreen button.

### DIFF
--- a/src/myanimelist/myanimelistClass.ts
+++ b/src/myanimelist/myanimelistClass.ts
@@ -6,51 +6,58 @@ import { activeLinks, removeFromOptions } from '../utils/quicklinksBuilder';
 import { waitForPageToBeVisible } from '../utils/general';
 
 export class MyAnimeListClass {
-  page: 'detail' | 'bookmarks' | 'modern' | 'classic' | 'character' | 'people' | 'search' | null =
-    null;
+  page:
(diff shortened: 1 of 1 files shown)


DFreezerR:
holy double quotes, my eslint is dying lmao
this code only works 1 time btw, can only do full screen on the first iframe opened.
consider updating attributes inside observer callback since it getting called on every DOM change, and you want to check and add attributes each time iframe gets created in DOM. I hope you got what I mean.

Source-Dom:
Oh damn you're right woops 😅 At least it's an ez fix, just have change the call order, also I'll fix the double quotes on the other parts 😂 

Thank you btw 🙏

henrik9999:
npm run lint:fix

NoPlagiarism:
> Don't merge yet

You can make it draft

Source-Dom:
Up

Source-Dom:
up

Source-Dom:
Up

Issue #2528, opened 2024-07-27
Display Broadcast in Local Timezone


bbb651:
### Suggestion

Currently broadcast time is always a string in the JST timezone like `Wednesday at 21:30 (JST)`.

I suggest:
- Displaying the timezone in the user's local timezone by default
- Wrapping the displayed value with the [\<time\>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time) element
- Have a setting to switch between `Local`, `JST` and `UTC` timezones

Which requires parsing the timezone from the provider.

PR #2654, opened 2024-10-23
Implement timezone configuration

halanm:
This PR implements a configuration section to change the broadcast timezone the default value is the user's current timezone.

Timezone settings page:

image

Old broadcast information string:

image

New broadcast information string in America/Vancouver timezone

image

New broadcast information string in America/Sao_Paulo timezone

image

This should close [#2528](https://github.com/MALSync/MALSync/issues/2528)

Let me know what you think about this and if everything is ok I'll work on adding the translations to other languages if it's needed


lolamtisch:
@halanm I have a few problems with this.

1. I am not actually sure what the use case is. I understand that you want that time in your own timezone. But is it really necessary to have an option? Is using the one set in your browser not enough, or do I miss somthing? I imagine it would be better to have a modal like with the different anime titles or so and display like the jp utc and your own, dont think more is necessary.

image

2. moment is such a bloated library, would be better to use browser tools instead if possible.

3. I do not like how the view is responsible for the transformation. I think the better way would be to extend the overview definition with a second body type inside of the info section. Then every provider (MAL, Anilist) has its own transformation to an date element. The moment it is a date it is easy to work with it. If it is not able to transform it it can return an fallback string. Then the view only has to format the date to the users timezone.
... | {
  date: date | string,
  type: weektime
} ...
https://github.com/MALSync/MALSync/blob/master/src/_provider/metaOverviewAbstract.ts#L22-L26

halanm:
I think showing all 3 of them (jst, utc and user's timezone) is a better approach. If done this way it's not necessary to have a timezone list and I can remove moment. I'll work on that when I have some time, thanks for suggesting

halanm:
@lolamtisch I don't know if that is exactly what you meant in the third suggestion. Please look it over and tell me if you think this can still be improved.

halanm:
I'll take a look

halanm:
Looks like this causes a lot of type errors like this one:

```
Property 'text' does not exist on type '{ text: string; url?: string | undefined; subtext?: string | undefined; } | { date: string | Date; type: "weektime"; }'.
  Property 'text' does not exist on type '{ date: string | Date; type: "weektime"; }'
 ```

halanm:
@lolamtisch What do you think? Should I fix the type errors or leave it as it is?

lolamtisch:
@halanm You are fine. I will take a look when I get to reviewing it.

lolamtisch:
@halanm I cleaned it up a bit. Could you review and test my changes?

halanm:
@lolamtisch Everything is looking good to me, thanks for that.

lolamtisch (approved):

Issue #2656, opened 2024-10-27
List Sync - Support syncing Start/Finish Date


Makar8000:
### Suggestion

In the current implementation, it seems that slaves have their start or finish date updated based on the current date when there's a status change. This results in the following issues:

- If a sync doesn't run until the day after the master was updated, the slave's date will be a day late.
- If the entry's dates were updated manually on the master to be at some point in the past (to make a correction or add an old entry), the dates will be incorrect

It seems that SImkl and Shikimori don't support start/finish dates, but it would be nice to support syncing for the ones that do. This would include syncing even when start or finish date are the only detected change.

Issue #800, opened 2021-07-21
Sync number of read manga volumes with List Sync


N0D4N:
Currently List Sync for manga syncs only read chapter count, score and status. It would be nice to have ability to sync read volumes count too, together with read chapter count.

PR #2657, opened 2024-10-28
List Sync improvements (Shikimori support, more fields, i18n)

Makar8000:
Implements #2656 and #800

This PR adds the following support to List Sync:
- Adds support for Shikimori
- Start/Finish Date for the providers that support it (AniList, MAL, and Kitsu)
  - Known Issue: Kitsu doesn't seem to like it when you try to set a start/finish date to null/empty. Sometimes it will update, but other times it will set it to the current date. To be more specific: if the master provider (not Kitsu) has a null start/finish date, but Kitsu (as a slave provider) has the start/finish date filled, then syncing will not behave properly for the date fields. I wasn't able to figure out a way to work around this. If desired, I can update this PR to not include this specific scenario in the diff list
- Manga Volumes
- Rewatch/Reread Count for the providers that support it (AniList, MAL, Kitsu, and Shikimori)
- Adjusts styling to highlight all changes instead of just changes to episode count
- Adjusts "EP" label to be "Ch" or "Ep" depending on if the item is Anime or Manga
- Adjusts the "Status" display to be the human-readable value rather than the integer value
- Fixes a bug where the rewatch count would not revert when clicking "Undo" after completing a rewatch for an entry
- i18n

Screenshot:
image
image


src/_provider/singleAbstract.ts
```diff
@@ -94,6 +101,60 @@ export abstract class SingleAbstract {
     return this._getStatus();
   }
 
+  abstract _setStartDate(startDate: definitions.startFinishDate): void;
+
+  public setStartDate(startDate: definitions.startFinishDate): SingleAbstract {
```
lolamtisch, line 106: startDate could be null too right?

src/_provider/singleAbstract.ts
```diff
@@ -94,6 +101,60 @@ export abstract class SingleAbstract {
...
+  public getStartDate(): definitions.startFinishDate | null {
+    if (this.supportsDates()) {
+      return this._getStartDate();
+    }
+    return null;
+  }
+
+  abstract _setFinishDate(finishDate: definitions.startFinishDate): void;
+
+  public setFinishDate(finishDate: definitions.startFinishDate): SingleAbstract {
```
lolamtisch, line 124: Same here could be null

src/_provider/Simkl/single.ts
```diff
@@ -56,12 +59,40 @@ export class Single extends SingleAbstract {
...
+  }
+
+  _setRewatchCount(rewatchCount) {
+    // Unsupported
+  }
+
+  _getRewatchCount() {
+    return null;
+  }
+
```
lolamtisch, line 95: Probably better to just throw an error unsupported. As we do not want the code to get to this point

src/_provider/MyAnimeList_api/single.ts
```diff
@@ -245,25 +280,6 @@ export class Single extends SingleAbstract {
   }
 
   async _sync() {
-    if (
```
lolamtisch, line 248: Needs discussion

lolamtisch (changes requested):

src/_provider/MyAnimeList_api/single.ts
```diff
@@ -72,6 +74,37 @@ export class Single extends SingleAbstract {
     }
   }
 
+  _getStartDate() {
+    return this.animeInfo.my_list_status.start_date;
```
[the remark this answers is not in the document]
Makar8000: How do you suggest these situations to be handled? Ideally it is always converted to `YYYY-MM-DD` format in `getStartDate()` and `getFinishDate()` so that other sites can parse it. Perhaps it could round down (`2022` -> `2022-01-01`, `2018-05` -> `2018-05-01`, etc)?

src/_provider/MyAnimeList_api/single.ts
```diff
@@ -72,6 +74,37 @@ export class Single extends SingleAbstract {
...
+  }
+
+  _getRewatchCount() {
+    if (this.type === 'manga') {
+      return this.animeInfo.my_list_status.num_times_reread;
+    }
+    return this.animeInfo.my_list_status.num_times_rewatched;
+  }
+
+  _setRewatchCount(rewatchCount) {
```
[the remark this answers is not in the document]
Makar8000: I could always just add `Math.min(rewatchCount, 255)`, but it wouldn't fix the list sync page always showing it in that case if MAL is set as a slave. Ex. if AniList as a master has 300 it will always say MAL is out of sync, but at least the sync itself will be successful

src/_provider/singleAbstract.ts
Makar8000: Resolved by adjusting `definitions.startFinishDate` to be `string | null` and removing the other `| null`

src/_provider/singleAbstract.ts
Makar8000: Resolved by adjusting `definitions.startFinishDate` to be `string | null` and removing the other `| null`

src/_provider/Simkl/single.ts
Makar8000: Fixed

src/_provider/MyAnimeList_api/single.ts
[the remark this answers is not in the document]
lolamtisch: Yeah I think that is the better way. Alternative would be returning null in theses cases, but that is not nice either.

I think rounding and returning null for the impossible combinations like year and day, in case they really exist. Is the way to go.

src/_provider/MyAnimeList_api/single.ts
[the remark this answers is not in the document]
Makar8000: Ah got it. I would agree then with not handling it

src/_provider/MyAnimeList_api/single.ts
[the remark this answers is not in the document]
Makar8000: Fixed

assets/_locales/en/messages.json
```diff
@@ -1061,6 +1061,74 @@
...
+    "settings_listsync_status": {
+        "message": "Status:",
+        "description": "List Sync label to display the user's watch/read status for an entry on a specific provider"
+    },
+    "settings_listsync_startdate": {
+        "message": "Start Date:",
+        "description": "List Sync label to display the user's start date for an entry on a specific provider"
+    },
+    "settings_listsync_finishdate": {
+        "message": "finish Date:",
```
lolamtisch, line 1117: Should probably be uppercase

lolamtisch (changes requested):

assets/_locales/en/messages.json
Makar8000: Oops. Good catch

lolamtisch (approved):

PR #2673, opened 2024-11-05
General fixes

DFreezerR:
Since there were some small changes, I decided not to split it into a bunch of small PRs.
Fixes:
- Missed language string not being translated 
- Local storage not being cleared when Clear Cache button was pressed in pop up extension (Settings -> Miscellaneous -> Clear Cache)
- Fixed Shikimori meta info - you could not get statistics info since it was filtered by only English selectors, requiring user to have English language selected on Shikimori website (API response you get based on language selected on Shikimori, so ENG = ENG API, RU = RU API).
- Fixed a bug when you change your locale (e.g. change browser UI language) and some strings in overview page of pop up extension were still in old language.
- Fixed 2 buttons in pop up being full width so you can easily mis-click them by accident ("Logged in via" button and "Log out" button)

I tested it on Firefox + Vivaldi (Chromium) and everything worked. The Local entries that were saved, are not being deleted by deleting local storage since the key for local entry is excluded from being cleared.
If you find something out of normal please tell, I'll check and change it.


src/utils/general.ts
```diff
@@ -891,11 +893,20 @@ export function waitForPageToBeVisible() {
   });
 }
 
+export function getBrowserCurrentLocale() {
```
lolamtisch, line 896: I am not sure if I want to use the complete browser locale for this.

I was planning to replace all the hour, mins etc with the Intl functions of the browser, so we do not have to do anything for the translation of time. But the problem I have is that I am not sure how to get the locale, the only way I could find is to creating an translation string for every language that contains the locale in it. A bit dirty but I do not see any cons in doing it like this. It would even correctly return it for the force english option. 

This would be a much better cache key indicator in this case and it would work well with the DurationFormat as locale too. What do you think about that?


https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat

src/utils/general.ts
```diff
@@ -891,11 +893,20 @@ export function waitForPageToBeVisible() {
...
+    return navigator.language;
+  }
+}
+
 export async function clearCache() {
   const cacheArray = await api.storage.list();
   let deleted = 0;
-
-  j.$.each(cacheArray, function (index, cache) {
+  localStore.clear();
```
lolamtisch, line 907: This is dangerous, because in the userscript it will clear the localStorage of the website. Which is not good. It can not be done in this utility as it can be executed in any context. It should only run in the extension and extension popup

src/_provider/metaOverviewAbstract.ts
```diff
@@ -109,17 +110,26 @@ export abstract class MetaOverviewAbstract {
   async init() {
     if (this.run) return this;
 
-    if (await this.getCache().hasValueAndIsNotEmpty()) {
+    const cache = await this.getCache();
```
lolamtisch, line 113: Does something speak against just using the locale inside of the cache key, which is defined at the end of this file?

assets/_locales/en/messages.json
```diff
@@ -1550,5 +1550,9 @@
     "settings_custom_position_below": {
         "message": "Below Information",
         "description": "Position of the quicklinks below the information section"
+    },
+    "settings_forceEn": {
```
lolamtisch, line 1554: I am always a bit reluctant in translating language switches. In case you do not understand the language it is in you can still find your way. But in this case it is less of a concern.

src/_provider/metaOverviewAbstract.ts
```diff
@@ -109,17 +110,26 @@ export abstract class MetaOverviewAbstract {
...
+        return this;
+      }
+      this.logger.log(`Locale changed [${data.locale} -> ${locale}], re-initializing...`);
+      await cache.clearValue();
     }
 
     await this._init();
     this.run = true;
-    this.getCache().setValue(this.getMeta());
+    this.getCache().setValue({ ...this.getMeta(), locale: getBrowserCurrentLocale() });
```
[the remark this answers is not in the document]
DFreezerR: Yes, I got what you mean. If user have like french lang in browser and force english in extension, they will have english UI but still french info + statistics. I'll make additional check for force english option

src/utils/general.ts
DFreezerR: To be honest I never touched this namespace, but looking into it, DurationFormat looks cool for durations I agree, but also [DateTimeFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat) I guess is nice, since US/Europe time formats.

But what you said about locale - aren't you already getting locale from browser with i18n inside `api.storage.lang`? I think it already checking whole browser lang when you call `chrome.i18n.getMessage`? 

And the reason I tried getting browser locale - to synchronize info and statistics fields inside meta overview when browser locale change. They need to be somehow handled, you can't just "update" them. the only thing I can think of is somehow evaluate their value on puling from local storage, or what I did. I don't know how to do the former one

src/utils/general.ts
DFreezerR: I thought when you use userscript and can only press this button from pwa page? Can it still clear website's localStorage too?

assets/_locales/en/messages.json
henrik9999: My opinion is also that it should be kept in english only

src/_provider/metaOverviewAbstract.ts
DFreezerR: It does not. I just thought it would be logical to put locale inside value rather than key, want me to change?

assets/_locales/en/messages.json
DFreezerR: I agree, if you don't know english enough to get what switch option tells you, you can't possibly understand the whole mess UI will turn after this switches. I'll revert this

src/utils/general.ts
lolamtisch: Malsync used to be a userscript only, without an extra domain and pwa. In the extension we removed the option to do it, but in the userscript you can still load the popup inside of the website. Many people switched to the userscript again just because of that feature.
image

src/utils/general.ts
lolamtisch: Ah I nearly forgot that DurationFormat is not supported in firefox. So we would need to implement an fallback for it until it is fully supported.

src/utils/general.ts
DFreezerR: Oh how to you open it? Didn't know, I'll add a check for an extension only

src/utils/general.ts
lolamtisch: > Ok let me show what I mean. I changed my browser lang from RU->ENG. The whole pop up UI lang also changed.

Ok I did get that. Probably just misinterpreted it and I do not mean not to use the locale. Just to use the malsync specif locale from the `chrome.i18n.getMessage` which would automatically fallback and support force english.

> No, I just assume they use the same function chrome.i18n.getUILanguage to get needed locale inside chrome.i18n.getMessage.  

I did not test it, but I imagine it returns the UI language of the browser, which can be different from the website prefered language (what the normal window js api returns) and again is different from the locale that malsync uses. But that is what I at least though. Never tested it.

> But they won't?... I don't get you, what part of the code you talking about? I changed some occurrences of "mins" "hours" etc, yes, but they pulled with api.storage.lang so they would be the same as UI. They already exist in messages.json.

Was probably not the best idea to mix the DurationFormat into the text like. I was talking about that and how you can pass the locale and that it would be a good key for the cache 

> You know, yes, I think it can solve getting locale. But one thing - how do you think it would work with Weblate? Can people mess with this string or you personally review translation PRs? If it's okay then I just add locale field in every locale file?

We review the changes. It opens an PR we have to accept.

> True just tested it. FF does have DateFormats btw. I can add them too if you ok?

Probably better in a seperate PR after this is done.

---

But genererally this is still in the discussion phase. So feel free to suggest alternatives if you have better. That is just what I was thinking about.

src/utils/general.ts
lolamtisch: As the idea of a util is that it can be used anywhere. I am not that confortable to include it in there.

I just found that we actually clear the localstorage when switching provider. It is probably a better idea to add it to the click event of the clear cache. This makes sure it runs in the context of the popup

https://github.com/MALSync/MALSync/blob/master/src/_minimal/components/settings/settings-structure-tracking.ts#L47

src/utils/general.ts
DFreezerR: >  it returns the UI language of the browser

True, this is browsers language. The preferred languages are in `navigator.language` (most preferred) and `navigator.languages` (all preferred)

> I was talking about that and how you can pass the locale and that it would be a good key for the cache

Got you, no problem

> discussion phase

To make it simpler for us, if you okay with this i just summarize:
1. Add locale translation key to every `message.json`
2. Pass it into Cache key parameter when it creates an entry
3. When getting overview meta from said cache, check current locale with cached one
3.1 If not equal - delete Cache entry and create new one
3.2 If equal - return it

I'm good with this as long as there are no "unexpected things" which I can't think of any right now.
And for DateFormat and DurationFormat I'll do another PR. How is it?

lolamtisch: Sounds good.

I think it would make sense to directly implement the durationformater, or at least directly use this utility for later. I think it pretty much does what durationformater does and could directly be used as fallback.

https://github.com/MALSync/MALSync/blob/master/src/utils/time.ts#L134

DFreezerR:
The only thing is - for old users, the locale check will be failed anyway since locale will not be in their old cache entries for the first time they open overview. But I guess that's not a big problem

src/_provider/metaOverviewAbstract.ts
```diff
@@ -165,11 +173,14 @@ export abstract class MetaOverviewAbstract {
...
   }
 
-  cacheObj: any = undefined;
+  cacheObj?: Cache = undefined;
 
   getCache() {
     if (this.cacheObj) return this.cacheObj;
-    this.cacheObj = new Cache(`v3/${this.url}`, 5 * 24 * 60 * 60 * 1000);
+    this.cacheObj = new Cache(
+      `v3/${api.storage.lang('locale')}/${this.url}}`,
```
lolamtisch, line 181: As the locale is in the cache key, meaning a different locale allways results in a new entry. This means that the comparision for a changed locale will never trigger. So the change could be removed again.

src/utils/Cache.ts
```diff
@@ -87,4 +87,8 @@ export class Cache {
     }
     return api.storage.get(this.key);
   }
+
+  getLocale() {
```
lolamtisch, line 91: Naming convention that is only true for metaOverviewAbstract, so it does not work for any other cache. As the locale is in the cache key the comparision in metaOverviewAbstract.ts is not necessary. So this function is not needed in the end and can be removed again.

src/utils/general.ts
```diff
@@ -894,8 +896,9 @@ export function waitForPageToBeVisible() {
 export async function clearCache() {
   const cacheArray = await api.storage.list();
   let deleted = 0;
+  if (api.type === 'webextension') localStore.clear();
```
lolamtisch, line 899: As this is a utility function. Meaning it can be used anywhere any time. I do not feel safe to clear the localstorage here.

I feel like putting it here is safer, as it makes it clearer that it should be executed in the context of the popup. While not totally safe, as the settings are build in a way that they could be injected anywhere. But still better than in the utility

https://github.com/MALSync/MALSync/blob/master/src/_minimal/components/settings/settings-structure-etc.ts#L44

src/utils/general.ts
```diff
@@ -894,8 +896,9 @@ export function waitForPageToBeVisible() {
...
 
-  j.$.each(cacheArray, function (index, cache) {
+  j.$.each(cacheArray, (index, cache) => {
```
lolamtisch, line 901: Just as an hint. This can be dangerous with jquery loops, as they often rely on `this`.

In this case it is fine. Specially as it make no sense that it is a jquery loop to beginn with instead of a normal js loop.

lolamtisch (changes requested):

src/_provider/metaOverviewAbstract.ts
DFreezerR: True, got up today and thought the same.

src/utils/Cache.ts
DFreezerR: Agree

src/utils/general.ts
DFreezerR: Yeah, now I think you are right, will fix it

src/utils/general.ts
DFreezerR: Thanks. When I first saw this I couldn't understand what it this loop

src/_minimal/components/settings/settings-structure-etc.ts
```diff
@@ -2,11 +2,12 @@ import { ConfObj } from '../../../_provider/definitions';
 import SettingsGeneral from './settings-general.vue';
 import SettingsGroup from './settings-group.vue';
 import SettingsClearTags from './settings-clear-tags.vue';
+import { localStore } from '../../../utils/localStore';
 
 export const etc: ConfObj[] = [
   {
     key: 'forceEn',
-    title: 'Force English',
```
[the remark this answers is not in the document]
DFreezerR: My bad, forgot to cherry pick that commit

src/_provider/metaOverviewAbstract.ts
```diff
@@ -171,11 +172,14 @@ export abstract class MetaOverviewAbstract {
...
   }
 
-  cacheObj: any = undefined;
+  cacheObj?: Cache = undefined;
 
   getCache() {
     if (this.cacheObj) return this.cacheObj;
-    this.cacheObj = new Cache(`v4/${this.url}`, 5 * 24 * 60 * 60 * 1000);
+    this.cacheObj = new Cache(
+      `v4/${api.storage.lang('locale')}/${this.url}}`,
```
lolamtisch, line 180: There is a } too much at the end

src/utils/general.ts
```diff
@@ -895,12 +895,12 @@ export async function clearCache() {
...
-  j.$.each(cacheArray, function (index, cache) {
-    if (!utils.syncRegex.test(String(index)) && !/(^tagSettings\/.*)/.test(String(index))) {
-      api.storage.remove(String(index));
+  for (let i = 0; i < cacheArray.length; i++) {
+    if (!utils.syncRegex.test(`${i}`) && !/(^tagSettings\/.*)/.test(`${i}`)) {
+      api.storage.remove(`${i}`);
       deleted++;
     }
-  });
+  }
```
lolamtisch, line 903: `api.storage.list()` apparently has the wrong type defintion it returns an object(`Promise<{ [key: string]: any }>`). Meaning it can not be handled like an array.

lolamtisch (changes requested):

src/_provider/metaOverviewAbstract.ts
DFreezerR: Somehow didn't see that..

src/utils/general.ts
DFreezerR: Always thought template string was faster than constructor or `toString()`. You against it?

> wrong type

I can change it

lolamtisch: You used object.keys on the wrong thing. it is still an index for loop and not a for of loop, meaning `i` is still only the index and not the object key.

I dont really care, but premature optimization is the root of all evil. ¯\\_(ツ)_/¯

DFreezerR: Oooh I got you, we need key not an index of it. Why I even thought about index. Guess my IQ is decreasing

DFreezerR: Done it, might've got too much locked on these `cacheArray` and `index` parameter in jquery loop lmao

PR #2683, opened 2024-11-10
Add Kavita Page

ZhiHanL:
Adding Kavita Support

### DIFF
--- a/src/pages/Kavita/main.ts
+++ b/src/pages/Kavita/main.ts
@@ -23,6 +23,9 @@ const seriesDetails = {
   seriesTitle: '',
   seriesID: '',
   totalPageNum: '',
+  anilistLink: '',
+  kitsuLink: '',
+  malLink: '',
 };
 
 async function getChapterProgress() {
@@ -65,12 +68,22 @@ export const Kavita: pageInterface = {
       return window.location.href.split('/').slice(0, 7).join('/');
     },
     getEpisode(url) {
-      if (!chapterDetails.chapterNum || !parseInt(chapterDetails.chapterNum))
+      if (
+        !chapterDetails.chapterNum ||
+        !parseInt(chapterDetails.chapterNum) ||
+        parseInt(chapterDetails.chapterNum) < 0
+      )
         throw 'No chapter number';
       return parseInt(chapterDetails.chapterNum);
     },
     getVolume(url) {
       return parseInt(chapterDetails.volumeNum);
+    },
+    getMalUrl(provider) {
+      if (provider === 'MAL' && seriesDetails.malLink) return seriesDetails.malLink;
+      if (provider === 'ANILIST' && seriesDetails.anilistLink) return seriesDetails.anilistLink;
+      if (provider === 'KITSU' && seriesDetails.kitsuLink) return seriesDetails.kitsuLink;
+      return false;
     },
     readerConfig: [
       {
@@ -112,7 +125,7 @@ export const Kavita: pageInterface = {
         return utils.urlStrip(window.location.href);
       },
     );
-    check();
+    waitForPageLoad(check);
 
     async function getSeriesDetails() {
       const seriesID = window.location.pathname.split('/')[4];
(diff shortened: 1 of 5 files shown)


src/pages/Kavita/main.ts
```diff
@@ -0,0 +1,176 @@
...
+      return seriesDetails.seriesTitle;
+    },
+    getIdentifier(url) {
+      if (!seriesDetails.seriesID) throw 'No series id';
+      return seriesDetails.seriesID;
+    },
+    getOverviewUrl(url) {
+      return window.location.href.split('/').slice(0, 7).join('/');
+    },
+    getEpisode(url) {
```
lolamtisch, line 67: For entries that do not have an chapterNum I get -100000

I think we should handle that

image

lolamtisch (changes requested):

src/pages/Kavita/main.ts
```diff
@@ -0,0 +1,176 @@
...
+          mode: 'callback',
+        },
+        total: {
+          callback: () => chapterDetails.totalPageNum as any,
+          mode: 'callback',
+        },
+      },
+    ],
+  },
+  overview: {
```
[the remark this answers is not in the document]
ZhiHanL: Done! What does it do exactly? I couldn't find anything about what providing these links do

src/pages/Kavita/style.less
```diff
@@ -0,0 +1,34 @@
...
+  select {
+    appearance: auto;
+    option {
+      background-color: var(--v-base-base);
+    }
+  }
+  input {
+    border-bottom: 1px solid;
+  }
+}
```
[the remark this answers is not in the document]
ZhiHanL: I think I like the way it looks as it is so I'll leave it

src/pages/Kavita/main.ts
```diff
@@ -0,0 +1,176 @@
...
+      clearInterval(progressUpdater);
+      if (!Kavita.isSyncPage(window.location.href) && !Kavita.isOverviewPage!(window.location.href))
+        return;
+
+      await getSeriesDetails();
+
+      if (Kavita.isSyncPage(window.location.href)) {
+        await getChapterDetails();
+        await getChapterProgress();
+        progressUpdater = setInterval(async () => {
```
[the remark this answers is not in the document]
ZhiHanL: Yeah, that's what I thought as well

src/pages/Kavita/main.ts
ZhiHanL: Done!

src/pages/Kavita/main.ts
```diff
@@ -0,0 +1,200 @@
...
+      const responseJSON = JSON.parse(response.responseText);
+      con.log('Chapter Info Retrieved', response);
+
+      chapterDetails.volumeNum = responseJSON.volumeNumber;
+      chapterDetails.chapterNum = responseJSON.chapterNumber < 0 ? 0 : responseJSON.chapterNumber;
+      chapterDetails.totalPageNum = responseJSON.pages;
+    }
+
+    function waitForPageLoad(callback) {
+      utils.waitUntilTrue(
```
lolamtisch, line 153: A bit dagerous. If the script loads late then it could miss it and then nothing happens.

It is porbably better to specially wait for that data. Like directly checking the title etc.

src/pages/Kavita/main.ts
```diff
@@ -0,0 +1,200 @@
...
+          );
+        },
+      );
+    }
+
+    function getLinks() {
+      seriesDetails.anilistLink =
+        j.$('[href^="https://anilist.co/manga/"]').attr('href') ?? seriesDetails.anilistLink;
+      seriesDetails.malLink =
+        j.$('[href^="https://myanimlist.net/manga/"]').attr('href') ?? seriesDetails.malLink;
```
DFreezerR, line 170: I'm not kavita user, but isn't that as typo? `https://myanimlist.net/manga/` -> `https://myanimelist.net/manga/`

src/pages/Kavita/main.ts
ZhiHanL: Ah that's true. The problem is that I added this because the mal link I'm searching for might not exist, so I can't just wait for that element.

I think I'll add a timeout to that page load, so if the spinner is missed, it'll still proceed with handling the page after 0.5s.

src/pages/Kavita/main.ts
```diff
@@ -0,0 +1,212 @@
...
+    },
+    getIdentifier(url) {
+      if (!seriesDetails.seriesID) throw 'No series id';
+      return seriesDetails.seriesID;
+    },
+    getOverviewUrl(url) {
+      return window.location.href.split('/').slice(0, 7).join('/');
+    },
+    getEpisode(url) {
+      if (!chapterDetails.chapterNum || !parseInt(chapterDetails.chapterNum))
```
lolamtisch, line 71: This condition triggers on chapterNum = 0. Which I think is not intended as it would mean it will never track volumen only entries.

lolamtisch (changes requested):

src/pages/Kavita/main.ts
ZhiHanL: Ah gotcha. I'll copy the Mangadex implementation: https://github.com/MALSync/MALSync/blob/master/src/pages/Mangadex/main.ts#L69

lolamtisch (approved):

PR #2687, opened 2024-11-12
settings: no disabling both types in quicklinks

NoPlagiarism:
firefox_MTGqHGgvgC


lolamtisch:
@NoPlagiarism Now that I think about it. The UI is not that well suited for this use case. Using a dropdown with `All` `manga` `anime` is probably better.

Something like the status selector or the filter button on the bookmarks. What do you think?

image
image

NoPlagiarism:
I am very sorry for missing this

> Using a dropdown with All manga anime is probably better.

I guess you're right, but I'm not good with Vue, so I'm not sure If I can make it atm. Maybe some-when later. Any contributions are appreciated tho

PR #2692, opened 2024-11-17
Add support for Disney Plus

NHClaessens:
I done my best to add support for Disney Plus, using the Jikan API to make up for any inconsistencies in how D+ represents it's data.

Some aspects are configurable, for which the settings can be found at the top of `src\pages\DisneyPlus\main.ts`.  
These settings are:
- How many Jikan entries to search for a matching episode name
- How similar the episode name must be for it to be considered a match (using Jaro-Winkler distance)
- Which type of title should be preferable returned: `English`, `Synonym`, `Default` or `Japanese`
- The minimum delay between two requests to Jikan


Source-Dom:
@lolamtisch just checking what's up with this PR

PR #2701, opened 2024-11-23
Anistream.moe

Flumuffel:


lolamtisch:
Is this page still alive?

NoPlagiarism:
Didn't they move to aninexus.to? I'm not sure, since didn't happen to witness .moe one

Flumuffel:
> Is this page still alive?

They moved the domain and branding to aninexus.to but the code didn't change from the exchange of data.

Issue #2710, opened 2024-11-27
FlameScans site update


SoldierZoom:
### Describe the Bug

The UI and URL structure has changed

### Steps to reproduce the behavior

go to https://flamecomics.xyz/

### Browser

Brave

### Type

Extension

### URL

https://flamecomics.xyz/

PR #2735, opened 2024-12-15
FlameScans integration updated

mega6382:
Solves #2710


src/pages/FlameScans/meta.json
```diff
@@ -1,6 +1,6 @@
 {
-  "search": "https://flamecomics.com/?s={searchtermPlus}",
+  "search": "https://flamecomics.xyz/browse?search={searchtermPlus}",
```
henrik9999, line 2: i think this has to be {searchterm}

src/pages/FlameScans/main.ts
```diff
@@ -2,24 +2,24 @@ import { pageInterface } from '../pageInterface';
...
-    return false;
+    return true;
   },
   sync: {
     getTitle(url) {
-      return j.$('.chapterbody .headpost a').text().trim();
+      return j.$('[class*="TopChapterNavbar_series_title"]').text().trim();
     },
     getIdentifier(url) {
       return replaceId(utils.urlPart(FlameScans.sync.getOverviewUrl(url), 4));
```
henrik9999, line 19: the replaceId function can probably be removed

haseeb-basil-bab:
@henrik9999 why is the test failing here? I ran the test locally and it worked just fine.

mega6382:
@henrik9999 why are the tests failing here? I ran them locally and they worked fine.

src/pages/FlameScans/main.ts
```diff
@@ -99,13 +100,16 @@ export const FlameScans: pageInterface = {
         con.error('404');
         return;
       }
-      if (j.$('h1.entry-title').length && j.$('.epcheck').length) {
+      if (
+        j.$('.mantine-Title-root').length &&
+        j.$('[class*="ChapterCard_chapterWrapper"]').length
```
lolamtisch, line 105: Is not part of the initial html source code, meaning it will not be triggered if the js loads it faster than malsycn can.

image

src/pages/FlameScans/main.ts
```diff
@@ -99,13 +100,16 @@ export const FlameScans: pageInterface = {
         con.error('404');
```
lolamtisch, line 100: As the page is a single page application now, meaning there is no reload between the pages. $.ready is not enough and It is necessary to listen for page changes and trigger handlePage again accordingly. 

Important navigations:
- Navigation to the overview page
- Navigation to the chapter page
- Chapter changes

lolamtisch (changes requested):

PR #2744, opened 2024-12-18
Update Voiranime domain

louisroyer:


henrik9999:
Maybe supporting *. Would be easier

henrik9999:
I will fix the test soon

PR #2750, opened 2024-12-24
New Page Q1N (Replaces Bakashi)

Martins-74:
The Bakashi page was permanently closed, and new page (created by an old admin of Bakashi) "Q1N" using the exact same schema and layout was put online. So i've made the necessary changes "removing" the old Bakashi and replace it the new Q1N.

P.s. bakashi.to it's not the same page as the older, it's a clone with different schema, that is the reason for me to replace it the Q1N.


Martins-74:
The failed tests is because this pages implements ddos-guard.

Issue #2716, opened 2024-12-05
[Website Request] Hikari


1MM0R7AL5:
### Website name

Hikari

### Website link

https://watch.hikaritv.xyz/

### Website language(s)

English

### Other details

This site lets you import MAL XML files, if that helps.

PR #2758, opened 2025-01-02
Add hikari site

Don-Burns:
Relating to https://github.com/MALSync/MALSync/issues/2716 adding support for [Hikari](https://watch.hikaritv.xyz/home)

I am a bit of a JS/TS noob so apologies in advance for obvious mistakes.

Tests not passing running locally and having a bit of trouble understanding how I should debug/test locally.
I am following https://github.com/MALSync/MALSync/wiki/Build docs which indicate to just open a PR sooner rather than later for help.

I used the [GoGoAnime](https://github.com/Don-Burns/MALSync/blob/master/src/pages/Gogoanime/main.ts) files as a guide as they seemed the most straightforward and similar to how Hikari works of the sites I personally know of already supported.

### DIFF
--- a/cspell.json
+++ b/cspell.json
@@ -10,6 +10,7 @@
         "Captcha",
         "eqeqeq",
         "greasemonkey",
+        "hikari",
         "itemprop",
         "jquery",
         "MALSync",
--- a/src/pages/Hikari/main.ts
+++ b/src/pages/Hikari/main.ts
@@ -1,0 +1,67 @@
+import { pageInterface } from '../pageInterface';
+
+export const Hikari: pageInterface = {
+  name: 'Hikari',
+  domain: ['https://watch.hikaritv.xyz'],
+  database: 'Hikari',
+  languages: ['English'],
+  type: 'anime',
+  isSyncPage(url) {
+    const isWatchPage = utils.urlPart(url, 3) === 'watch';
+    // sync page has an `eps` parameter denoting the current episode
+    const hasEpNumber = utils.urlParam(url, 'eps') !== null;
+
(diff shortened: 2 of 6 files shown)


henrik9999:
You need to add it in https://github.com/MALSync/MALSync/blob/master/src/pages/pages.ts

Don-Burns:
> You need to add it in https://github.com/MALSync/MALSync/blob/master/src/pages/pages.ts

🤦 skimmed right over that in the docs. Thanks!

Should be working now, when installing locally the UI elements are showing up as expected.

lolamtisch:
Unfortunately is the design to close to hianime. Making it difficult to not see it as a clone, even though the content is not.

But apparently are they going to revise the UI soon. I will put this PR on ice until that happens. Hopefully then it is not that big of a problem anymore.

Don-Burns:
Hi, fair enough @lolamtisch.  Will admit I didn't know about hianime, they do look very similar so I can get that you don't want to support a UI clone.

tigeryu8900:
The site had an UI update.

Don-Burns:
I squashed my fixup commit to clean up the commit history since we were just waiting for the UI update (which appears to have been done) and to retrigger CI.
From what I can see locally this is still working the same as before without any modification needed.

lolamtisch:
@Don-Burns Sorry we already did an alternative implementation of the page. I should have informed you. But feel free to give the other implemation an test run.

https://github.com/MALSync/MALSync/pull/2826

Don-Burns:
Sure no worries, I'll close this so.

PR #2795, opened 2025-01-20
Fix Crunchyroll overview page selector

### LINKED ISSUE #2788: Overview page not identified after Crunchyroll site update
### Describe the Bug

After Crunchyroll updated their site a few days ago, the episode list page ("series" page) is no longer identified by MALSync.  

Looks like they removed `.erc-series-hero` element from the page, which is used by MALSync to identify the overview page, see CR's `isOverviewPage()`:

https://github.com/MALSync/MALSync/blob/07f04fcd918eed1bc3067e2d78ddf0642d686bb2/src/pages/Crunchyroll/beta.ts#L91-L93

Open any episode list page on crunchyroll.
The correction popup icon and the `#malp` menu are gone. Clicking extension no longer automatically identifies the series.

### Browser

Chrome 132.0.6834.84 (Official Build) (x86_64)

### Type

Extension

### URL

https://www.crunchyroll.com/series/GG5H5XQX4/frieren-beyond-journeys-end

sergiitk:
Fixes #2788.

Tested with locally built extension:
Screenshot 2025-01-19 at 4 06 37 PM


src/pages/Crunchyroll/beta.ts
```diff
@@ -89,7 +89,7 @@ export const beta: pageInterface = {
     return Boolean(j.$('.erc-watch-episode-layout').length);
   },
   isOverviewPage(url) {
-    return Boolean(j.$('.erc-series-hero').length);
```
lolamtisch, line 92: We specially did not want to use the url for the check, because crunchyroll does strange things with its urls sometimes. They used to translate the url paths, which currently seems like not beeing the case anymore. But we wanted to be safe. But apparently this backfired. But you are right, now it is probably stable enough to use the url.

Still this change will only work for the US version of the website as the other languages have a locale prefix in the url. 
https://www.crunchyroll.com/de/series/G4PH0WJQZ/fatestrange-fake

lolamtisch (changes requested):

src/pages/Crunchyroll/beta.ts
sergiitk: Ah, I was wondering why this DOM selector and not a simple url check. Let me see if I can find another selector to hook up to.

sergiitk: Done, replaced with `.erc-season-episode-list`.

sergiitk: The German page you link works now too.

lolamtisch (approved):

PR #2808, opened 2025-01-26
Fallback to other sync modes

tigeryu8900:
I added the option `syncFallback` to use other sync modes for entries not available for the current sync mode.

image

On this [MangaDex page](https://mangadex.org/title/3b62f955-732c-43b2-84e7-cc1ff57896a7/abandoned-tsuyosugite-buki-ga-kowareru-yuusha-to-buki-shokunin-no-elf) when `syncMode` is `ANILIST` and `syncFallback` is `false`, MALSync falls back to `Local`.

image

When `syncFallback` is `true`, MALSync falls back to `MAL` instead.

image

The entry also shows up in user lists, though it requires scrolling past all the `mode == syncMode` entries (maybe find a way to merge the user lists?). Duplicate entries are removed by comparing `cacheKey`.

image

Search works by concatenating the search from different modes with `mode == syncMode` entries at the top, but duplicates aren't removed since there isn't a simple way to remove them.

image

### DIFF
--- a/assets/_locales/de/messages.json
+++ b/assets/_locales/de/messages.json
@@ -627,6 +627,10 @@
         "message": "Modus",
         "description": "Mode"
     },
+    "settings_Fallback": {
+        "message": "Fallback to other modes",
+        "description": "Fallback to other modes"
+    },
     "settings_Authenticate": {
         "message": "Authentifizieren",
         "description": "Authenticate"
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -823,6 +823,10 @@
         "message": "Mode",
         "description": "Mode"
     },
+    "settings_Fallback": {
+        "message": "Fallback to other modes",
+        "description": "Fallback to other modes"
+    },
     "settings_Authenticate": {
         "message": "Authenticate",
         "description": "Authenticate"
--- a/assets/_locales/es/messages.json
+++ b/assets/_locales/es/messages.json
@@ -627,6 +627,10 @@
         "message": "Modo",
         "description": "Mode"
     },
+    "settings_Fallback": {
+        "message": "Fallback to other modes",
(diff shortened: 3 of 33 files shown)


lolamtisch:
@tigeryu8900 We always where a bit against this kind of an implementation. As it makes the updates a bit intransparent. While the mixing of the lists helps, it is error prone and makes paginated loading not viable. Specially because the different pages entries do not always have an matching counter part.

We had an alternative idea during internal discussion which I think should solve some of the problem here. Our idea was to use the current local-sync entries as some kind of proxy entry. 

Lifecycle:
1. Click on "not on mal" on a website
2. You will be able to somehow save the alternative tracker url in the local-entry. Not sure how this will be UI wise yet. We can think about that later.
3. Every change to the local-entry will be mirrored to the alternative tracker.
4. Opening an local-entry will always get the data from the alternative tracker as state.

Pros:
- Does allow as many alternative trackers as one has, while still showing up in the list of the popup.

Drawbacks: 
- Lists will not show updates done on the alternative trackers directly, as then the local entry will be out of sync. But opening the local entry, will always trigger a sync of the remote to the local entry.
- Local entries do not sync between different browser vendors.


What do you think of this idea? We can discuss it further on discord if you want to.

lolamtisch:
1. The id needs to stay the same as we use the ID to identify which entry should be local, incase you delete your caches it will auto map it again.

2. I rather have entries that are out of sync, than so many extra requets. I am not a fan of mixing sources. Only local storage should be allowed to be mixed, as it does not require extra requests.
List -> primary tracker + local storage
Normal entry detail page -> primary tracker
local entry detail page -> local storage + alternative tracker.

3. The problem is how to detect that enties finally exist on the primary tracker.

4. We only have quite limited Extension storage. So this is not an option.

PR #2829, opened 2025-02-12
Shikimori - GraphQL

DFreezerR:
Partially moved Shikimori to GRAPHQL API

Problem: While Shikimori uses old REST API, there's no way to user to search for LNs/WNs in correction search or normal search using Shiki as primary provider. That is due to manga and novels being separate API endpoints while other providers (MAL, AL etc.) merged them together under name "Manga". 

Updating to GQL allow to search for both manga types and novel types at the same time since their API merged novels with manga.

Implementation modifies almost all Shikimori files, so there's a need for writing a tests (I think).

Some notable changes:

1. GQL impl.
2. Split API into "queries" and "types"
3. Old REST API only left for adding/deleting/updating user list
4. Added "updated_at" sorting option to sort user list by update time in both orders asc/desc

Also some spelling fixes.

### DIFF
--- a/src/_provider/Shikimori/single.ts
+++ b/src/_provider/Shikimori/single.ts
@@ -215,13 +215,10 @@ export class Single extends SingleAbstract {
   }
 
   async _sync() {
-    if (!this.userRate) {
-      return;
-    }
     if (this._onList) {
-      await Queries.UserRateUpdate(this.userRate);
+      await Queries.UserRateUpdate(this.userRate!);
     } else {
-      await Queries.UserRateAdd(this.userRate);
+      await Queries.UserRateAdd(this.userRate!);
     }
   }
 
@@ -230,9 +227,6 @@ export class Single extends SingleAbstract {
   }
 
   async _delete() {
-    if (!this.userRate) {
-      return;
-    }
-    await Queries.UserRateDelete(this.userRate.id);
+    await Queries.UserRateDelete(this.userRate!.id);
   }
 }


DFreezerR:
I think it's safe to review now

src/_provider/Shikimori/single.ts
```diff
@@ -46,11 +51,12 @@ export class Single extends SingleAbstract {
...
 
   _getStatus() {
-    return helper.statusTranslate[this.animeInfo!.status];
+    return Number(statusTranslate[this.userRate!.status]);
   }
 
-  _setStatus(status) {
-    this.animeInfo!.status = helper.statusTranslate[status] as helper.StatusType;
+  _setStatus(status: definitions.status) {
+    if (!this.userRate) return;
```
lolamtisch: Should `this.userRate` not always be populated after a `_update`? In that case this check is unnecessary. The only case where this does something is when `_update` is not executed, which should generally never happen. It is even harmfull as it would hide the problem. I imagine the solution is either remove or throw an error.

This applies to all the other functions too. And I think `this.metaInfo` should be the same then

src/_provider/Shikimori/helper.ts
```diff
@@ -1,6 +1,14 @@
-import { NotAutenticatedError, NotFoundError, ServerOfflineError } from '../Errors';
-import { status } from '../definitions';
+/* eslint-disable import/no-cycle */
```
lolamtisch: Import cycles are often unpredictable. Please fix the cycle instead of silenting it.

src/_provider/Shikimori/helper.ts
```diff
@@ -14,60 +22,57 @@ export async function apiCall(options: {
...
       headers,
-      data,
+      data: options.dataObj ? JSON.stringify(options.dataObj) : undefined,
     })
     .then(async response => {
       if ((response.status > 499 && response.status < 600) || response.status === 0) {
         throw new ServerOfflineError(`Server Offline status: ${response.status}`);
       }
 
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
lolamtisch: Same here is it not just a warning?

src/_provider/Shikimori/helper.ts
```diff
@@ -135,32 +158,22 @@ async function refreshToken(refresh_token: string) {
...
-      });
-    }
-    return res;
-  });
+export async function userRequest() {
+  return Queries.CurrentUser();
 }
 
-export async function userId() {
+export async function userId(): Promise<string> {
```
lolamtisch: It would be better to do, this way we do not directly need to define the return type.
```return cacheObj.getValue() as Promise<string>;```

or extend the cache class to take generics.
```Cache<string>```

File: src/_provider/Shikimori/helper.ts [hunk 5]
```diff
@@ -14,60 +22,57 @@ export async function apiCall(options: {
   type: 'GET' | 'PUT' | 'DELETE' | 'POST';
   path: string;
   parameter?: { [key: string]: string | number };
-  dataObj?: { [key: string]: any };
+  dataObj?: { [key: string]: unknown };
   auth?: boolean;
-}) {
-  const type = options.type || 'GET';
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
lolamtisch: Is it not just a warning? In that case the disabling of the rule is not necessary. Kind of "dangerous" for example if we decide to change no-explicit-any from warning to error in the future.

src/_provider/Shikimori/list.ts
```diff
@@ -1,23 +1,40 @@
-import { ListAbstract, listElement } from '../listAbstract';
+/* eslint-disable no-await-in-loop */
```
lolamtisch: Remove as only a warning

src/_provider/Shikimori/list.ts
```diff
@@ -26,110 +43,91 @@ export class UserList extends ListAbstract {
...
-        rewatchCount: entry.rewatches,
-        image: meta.image.original ? `${helper.domain}${meta.image.original}` : '',
-        imageLarge: meta.image.original ? `${helper.domain}${meta.image.original}` : '',
-        tags: entry.text,
+        malId: item.malId,
+        apiCacheKey: item.id,
+        uid: item.id,
+        cacheKey: item.id,
+        type: this.listType,
+        title: helper.title(item.russian || '', item.english || item.name),
```
lolamtisch, line 119: You forgot to get the name in the list query. Resulting in an empty title
image

src/_provider/Shikimori/queries.ts
```diff
@@ -0,0 +1,462 @@
...
+  static CurrentUser(): Promise<Types.CurrentUser> {
+    return apiCall({
+      type: 'POST',
+      path: 'graphql',
+      dataObj: {
+        query: GRAPHQL.CurrentUser,
+      },
+    }).then((res: Types.CurrentUser) => {
+      // TODO - Waiting for GRAPHQL update
+      // if (res.locale) {
```
lolamtisch: This is unfortunate. As this will result in mixed languages in the popup. RU title with English descriptions and headers. I am fine with it tho, as I imagine most people have Shiki set to RU either way
image

lolamtisch (changes requested):

src/_provider/Shikimori/single.ts
DFreezerR: Oh yes, I think I did all of these checks just to get rid of the TS warning that variable might be undefined. I look into it once again

src/_provider/Shikimori/helper.ts
DFreezerR: Fixed by moving to queries

src/_provider/Shikimori/helper.ts
DFreezerR: Removed (I just hate yellow warning lines)

src/_provider/Shikimori/helper.ts
DFreezerR: Extended

src/_provider/Shikimori/helper.ts
DFreezerR: Removed x2

src/_provider/Shikimori/list.ts
DFreezerR: Removed x3

src/_provider/Shikimori/list.ts
DFreezerR: Fixed. Idk how I forgot a field in GQL

src/_provider/Shikimori/queries.ts
DFreezerR: I added old request back, can you try this one again?

PR #2852, opened 2025-02-23
HiAnime fix ui selector

tigeryu8900:
Changes the location of MAL-Sync's UI on HiAnime to be consistent with other sites, such as MangaDex, AnimeKAI, etc.

Also adds MAL-Sync's UI on the sync page.

### DIFF
--- a/src/pages/Zoro/main.ts
+++ b/src/pages/Zoro/main.ts
@@ -42,6 +42,9 @@ export const Zoro: pageInterface = {
     nextEpUrl(url) {
       return jsonData.next_episode_url;
     },
+    uiSelector(selector) {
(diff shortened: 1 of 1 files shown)


src/pages/Zoro/main.ts
```diff
@@ -42,6 +42,9 @@ export const Zoro: pageInterface = {
     nextEpUrl(url) {
       return jsonData.next_episode_url;
     },
+    uiSelector(selector) {
```
henrik9999, line 45: We dont add the ui on episodes by choice if they have a overview so Plesse remove this

tigeryu8900: Ok, removed

PR #2878, opened 2025-03-12
Chibiscript

lolamtisch:


src/background/customDomain.ts
```diff
@@ -27,7 +28,16 @@ async function registerScripts() {
     return;
   }
 
-  const domains: domainType[] = await api.settings.getAsync('customDomains');
+  let domains: domainType[] = await api.settings.getAsync('customDomains');
```
Source-Dom, line 31: Does it not work anymore as a constant?

src/background/customDomain.ts
```diff
@@ -27,7 +28,16 @@ async function registerScripts() {
...
+
+  try {
+    const chibiRepo = await ChibiListRepository.getInstance().init();
+    const chibiDomains = chibiRepo.getPermissions();
+    domains = domains.concat(chibiDomains);
```
Source-Dom, line 36: Can you do that without `domains = ` or it does not work ?

src/chibiScript/functions/controlFlowFunctions.ts
```diff
@@ -0,0 +1,52 @@
...
+    condition: ChibiJson<boolean>,
+    thenAction: Args[1],
+    elseAction: Args[2],
+  ): CT.UnwrapJson<Args[1]> | CT.UnwrapJson<Args[2]> {
+    const conditionState = ctx.run(condition);
+    if (conditionState) {
+      return ctx.run(thenAction) as CT.UnwrapJson<Args[1]>;
+    }
+    return ctx.run(elseAction) as CT.UnwrapJson<Args[2]>;
+  },
```
DFreezerR, line 32: I'm not good with TS typing but for me `if` returns `ChibiJson<unknown>`  while for example nextEpUrl should return `ChibiJson<string | boolean | undefined>` and because of that error happens. Idk if this is intended. It's possible to cast it with `$c.if(...).string()` or `$c.if<ChibiJson<boolean>, ChibiJson<string>, ChibiJson<string>>(...)` but it looks just weird

webpackConfig/plugins/ExtractJsonPlugin.js
DFreezerR, line 1: Just like @tigeryu8900 said in discord, I think `webpackConfig\plugins\ExtractJsonPlugin.js` should be fixed (also it's import in `webpackConfig\webextension.content.config.js`), since I can not build it like this. Can include fix into one of PRs if you want.

src/chibiScript/functions/core/coreFunctions.ts
```diff
@@ -0,0 +1,267 @@
...
+   */
+  provider: (ctx: ChibiCtx, input: void): 'MAL' | 'ANILIST' | 'KITSU' | 'SIMKL' => {
+    const provider = ctx.get('provider');
+
+    if (!provider) {
+      throw new ChibiError('Provider not set');
+    }
+
+    return provider;
+  },
```
DFreezerR, line 39: Shouldn't provider function also include `SHIKI` and `MALAPI` as possible return values?

src/chibiScript/functions/objectFunctions.ts
DFreezerR, line 1: I may be blind but there's no `object` function to create chibi objects? Like with string, boolean, array and number. It's even in object function descriptions `$c.object({user: {name: "John"}}).get("user") // returns {name: "John"}` in https://github.com/MALSync/MALSync/pull/2878/files#diff-e69a5cc7793edb4456c21b36b0e702f748550698655a31ab56c9a0af7e9cb845R4-R11

src/chibiScript/functions/conditionFunctions.ts
```diff
@@ -0,0 +1,157 @@
...
+   * @input void - No input used
+   * @param values - One or more boolean values to OR together
+   * @returns Result of OR operation on all values
+   * @example
+   * $c.or($c.boolean(false).run(), $c.boolean(true).run()) // returns true
+   * $c.or($c.boolean(false).run(), $c.boolean(false).run()) // returns false
+   */
+  or: (ctx: ChibiCtx, input: void, ...values: ChibiJson<boolean>[]) => {
+    return values.reduce((result, val) => result || ctx.run(val), false);
+  },
```
DFreezerR, line 115: What you think about allowing `or` to be used with any type to imitate `||` operator for non-booleans? For example:
`$c.or(title1, title2).run()` will return title1 if its truthy, and title2 otherwise and so on? It's already kinda possible and it's working but there are a lot of typing errors/warnings. If you think that's too much than it's ok

DFreezerR (commented):
That's all I remembered for now. If find something else I'll tell

src/chibiScript/functions/objectFunctions.ts
lolamtisch: Yeah I know. I was not sure what the usecase would be. But I can add it, as it does not hurt. In the end one could do the same with string('{...}').jsonParse();

src/pages-chibi/pageInterface.ts
```diff
@@ -0,0 +1,303 @@
...
+     */
+    getMalUrl?: (
+      $c: ChibiGenerator<unknown>,
+    ) => ChibiJson<Promise<string | boolean> | string | boolean>;
+
+    /**
+     * Configuration for the manga progress functionality.
+     * This property is optional.
+     */
+    readerConfig?: mangaProgressConfig[];
```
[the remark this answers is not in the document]
lolamtisch: I am still not sure what to do with readerConfig. I think my plan for the future is to completly migrate it to chibiscript, as having 2 different json systems is a bit too much. But for now I just did not want to put more work into it, until I know that the extension stores accept the chibi system. I will handle until now that I add readerConfig myself.

src/background/customDomain.ts
lolamtisch: I just want the custom domains continue to work in case the chibi implementation throws an error.

webpackConfig/plugins/ExtractJsonPlugin.js
lolamtisch: What exactly is the problem?

src/background/customDomain.ts
Source-Dom: So are you saying that doing `domains.concat(chibiDomains);` without `domains =` wouldn't work ?

I'm not understanding why it wouldn't work, the error would be thrown before concatenating so it should be fine, am I wrong ?
Also the previous "customDomains" would still be present because they are added way before when creating the constant ? 

Am I missing something ?

src/pages-chibi/pageInterface.ts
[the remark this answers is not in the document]
DFreezerR: Yeah I also wanted to suggest rewriting it into chibi completely but thought it might be too much
All good then

webpackConfig/plugins/ExtractJsonPlugin.js
DFreezerR: The entire project does not build, webpack throws error, I mentioned it in discord
https://discord.com/channels/358599430502481920/410081383482916864/1352644356300865687

lolamtisch: Hmm I do not get that. What nodejs version do you use?

DFreezerR: I have 
```
❯ node -v
v20.15.1
```
installed which should satisfy node version in `package.json`

src/chibiScript/functions/objectFunctions.ts
DFreezerR: It just was the first thought I had when I saw functions for objects. Just thinking it would be logical to have it

src/background/customDomain.ts
Source-Dom: > > So are you saying that doing `domains.concat(chibiDomains);` without `domains =` wouldn't work ?
> > I'm not understanding why it wouldn't work, the error would be thrown before concatenating so it should be fine, am I wrong ? Also the previous "customDomains" would still be present because they are added way before when creating the constant ?
> > Am I missing something ?
> > _Sorry for focusing on this little detail 😅_
> 
> https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/concat `concat()` does not modify existing arrays, it returns combined one, so to get both normal domains and chibi ones you have to assign result of `concat()` somewhere

Ohhh damn my bad x) Thank you lol

Source-Dom: I'll delete this then :)

src/chibiScript/functions/conditionFunctions.ts
tigeryu8900: I think we should create a separate function so it doesn't mess with places where `ChibiJson<boolean>` are used. Maybe something like this:

```typescript
$c.coalesce(title1, title2).run()
```

Kind of like [MySQL's `COALESCE` function](https://www.w3schools.com/sql/func_mysql_coalesce.asp).

webpackConfig/plugins/ExtractJsonPlugin.js
lolamtisch: Added 20.19 as requirement

src/chibiScript/functions/core/coreFunctions.ts
lolamtisch: Added

src/chibiScript/functions/objectFunctions.ts
lolamtisch: Added

src/chibiScript/functions/conditionFunctions.ts
lolamtisch: Added as coalesce

DFreezerR:
LGTM👍

src/chibiScript/functions/core/coreFunctions.ts
```diff
@@ -0,0 +1,267 @@
...
+
+  /**
+   * Gets the current element from context
+   * @returns Current element
+   * @example
+   * $c.element()
+   */
+  element: (ctx: ChibiCtx, input: void): Element => {
```
tigeryu8900, line 47: This might be confused with literal functions. Another name might be better, like `$0` which dev console uses to refers to the currently selected element, or `target` which refers to the target element of a dom event.

We could also add the literal function `element` since we might want to cast a `ChibiJson<any>` to `ChibiJson<Element>` or create a `ChibiJson<Element>` from a string.

src/chibiScript/functions/stringFunctions.ts
```diff
@@ -0,0 +1,202 @@
...
+  /**
+   * Concatenates two strings
+   * @input string - The first string
+   * @param value - The second string to concatenate
+   * @returns Concatenated string
+   * @example
+   * $c.string("Hello").concat(" World") // returns "Hello World"
+   */
+  concat: (ctx: ChibiCtx, input: string, value: string) => {
+    return input.concat(value);
```
tigeryu8900, line 189: We should also support `ChibiJson<string>` for `value`.

PR #2916, opened 2025-04-15
add types to console and remove consoleBG

PatrickMatthiesen:
I noticed that we had an entire extra console log file just to distinguish the service worker from the rest of the logs.
I have made it a variable in webpack so we can resolve it at build time and change the text that way.

I also added typescript types to the file and fixed some names to remove linter errors.


PatrickMatthiesen:
@lolamtisch Anything that I need to do to get this approved?

PR #2926, opened 2025-04-16
Fix linter warnings and refactor

PatrickMatthiesen:
A bunch of linter fixes.

I renamed variables, added words to the dictionary, added missing await, and more to come.

Please give me feedback if you have any
Especially if I'm breaking something that should have backwards compatibility 🙂


PatrickMatthiesen:
This might have gotten a bit out of hand with the number of files, but most of them are, luckily, just because I renamed things.
I have tried to break up most of the changes into different types of changes in each commit.

There are a few files that have larger changes than expected because of inconsistencies in the use of the methods. The inconsistencies made it hard to type, so I made slight changes to align the usage. I also did my best to debug the code when possible, to make sure that the runtime fit the static types.

I also seem to be having some issues with a missing quicklinks.json file that I can see is being created with build:extension. I will wait for the Chibiscript branch to merge into master before I making attempts at fixing the issue.

PatrickMatthiesen:
For the page tests
- https://kisshentai.net/ does not exist anymore
- Animelon fails the same way on master
- animepahe: page.addScriptTag adds the script, but it does not seem to execute it ¯\_(ツ)_/¯
 - fails on master too
- TRanimeizle cant test because I'm blocked 😅

The normal unit test seems to fail in other pull requests too.

PatrickMatthiesen:
Should I pull out the rename of pageInterface -> PageInterface to another PR so I can reduce the number of files this PR affects?

PR #2947, opened 2025-04-18
Update Toonily.com to new url path

LoLsAlEd:
I noticed that [Toonily.com](https://toonily.com/) was broken and the issue appears to stem from the URL path for the [webtoons](https://toonily.com/serie/overgeared/) changing from `webtoon` to `serie` and updating this appears to fix support for the site as both the integration on the pages and search functionality appear to be working again when loading the extension from disk. 

Please let me know if there are any issues and I will try to address them.

Thanks!


henrik9999:
Has been done in the mentioned commit

PR #2949, opened 2025-04-22
ChibiScript - fix types

tigeryu8900:


src/chibiScript/ChibiRegistry.ts
```diff
@@ -1,4 +1,12 @@
-export const reservedKeys = ['url', 'ui', 'provider', 'trigger', 'pageObject', 'element'];
+export type ReservedKey = 'url' | 'ui' | 'provider' | 'trigger' | 'pageObject' | 'element';
+export const reservedKeys: ReservedKey[] = [
+  'url',
+  'ui',
+  'provider',
+  'trigger',
+  'pageObject',
+  'element',
+];
```
lolamtisch, line 9: I think this should be the same just without double definitions
```typescript
export const reservedKeys = ['url', 'ui', 'provider', 'trigger', 'pageObject', 'element'] as const;
export type ReservedKey = (typeof reservedKeys)[number];
```

src/chibiScript/functions/core/coreFunctions.ts
```diff
@@ -61,7 +61,7 @@ export default {
    * @example
    * $c.string('/anime/123').ifThen($c => $c.urlAbsolute().return().run()).boolean(false)
    */
-  return: (ctx: ChibiCtx, input: any): void => {
+  return: <T = any>(ctx: ChibiCtx, input: T): void => {
```
lolamtisch, line 64: I do not understand what exactly this does? For input it often makes more sense to use `<input> (ctx, input: Input)` as then it uses the real input of the class. https://github.com/MALSync/MALSync/blob/3810e48b69234ff2a938b5d9b3798ef826550664/webpackConfig/webextension.content.generateChibi.js#L38

src/chibiScript/functions/core/coreFunctions.ts
```diff
@@ -143,7 +173,7 @@ export default {
    * @example
    * $c.getVariable('myVar', 'default') // returns the value of myVar or 'default' if not set
    */
-  getVariable: (ctx: ChibiCtx, input: void, key: string, defaultValue?: any): any => {
+  getVariable: <T = any>(ctx: ChibiCtx, input: void, key: string, defaultValue?: T): T => {
```
lolamtisch, line 176: I am not sure if we really should use the defaultValue type as return type. It is more of type `: any | T` which in the end is any

src/chibiScript/functions/core/coreFunctions.ts
```diff
@@ -181,7 +216,7 @@ export default {
    * @example
    * $c.getGlobalVariable('myGlobalVar', 'default') // returns the value of myGlobalVar or 'default' if not set
    */
-  getGlobalVariable: (ctx: ChibiCtx, input: void, key: string, defaultValue?: any): any => {
+  getGlobalVariable: <T = any>(ctx: ChibiCtx, input: void, key: string, defaultValue?: T): T => {
```
lolamtisch, line 219: Same note as above

src/chibiScript/functions/core/coreFunctions.ts
```diff
@@ -87,10 +87,40 @@ export default {
    * @example
    * $c.url().this('overview.getIdentifier') // Calls the getTitle function with the current URL as parameter
    */
-  this: (ctx: ChibiCtx, input: any, property: string): any => {
+  this: <Property extends string>(
+    ctx: ChibiCtx,
+    input: any,
+    property: Property,
+  ): Property extends `${string}.is${string}` | `is${string}`
```
lolamtisch, line 94: Works for now. In the future we could think about building the types from an union of both pageInterfaces.

lolamtisch (changes requested):
Thank you. Some good finds

src/chibiScript/ChibiRegistry.ts
tigeryu8900: Fixed. I also added the `isReservedKey` function:

```typescript
export function isReservedKey<T extends string>(key: T): T extends ReservedKey ? true : false {
  return reservedKeys.includes(key as ReservedKey) as ReturnType<typeof isReservedKey<T>>;
}
```

src/chibiScript/functions/core/coreFunctions.ts
tigeryu8900: Got it. I reverted this change.

src/chibiScript/functions/core/coreFunctions.ts
tigeryu8900: Got it. I reverted this change.

src/chibiScript/functions/core/coreFunctions.ts
tigeryu8900: Got it. I reverted this change.

src/chibiScript/functions/core/coreFunctions.ts
tigeryu8900: I think we could generate the function body as well, not just the types.

lolamtisch: Yeah but it is good enough for now

PR #2976, opened 2025-05-05
add anisearch

Stinkerwue:


lolamtisch:
Can you even watch anime on anisearch? But would it not be kind of strange to watch anime on an anime db page only to sync it to a different db page?

Stinkerwue:
Then i missunderstand the wiki.

Issue #3009, opened 2025-05-21
[Website Request] AnimeAV1


SaraLunaDev:
### Website name

AnimeAV1

### Website link

https://animeav1.com/

### Website language(s)

Spanish

DFreezerR:
Done #3038

SaraLunaDev:
Maybe i'm asking too much but would be great to fit the MAL info next to the last div instead of above. This would avoid the weird under space the cover has right now. And if it's easy to add, rounded borders to the last episode seen to match the aesthetic of the web would be amazing too! 

[Reference](https://i.imgur.com/1bQehJi.png)

henrik9999:
@dolvem

dolvem:
Fixed in the PR #3065, this is the result: [Reference](https://imgur.com/a/kYuNp8h)

PR #3036, opened 2025-06-07
Add Animeav1 support [ES]

dolvem:
This PR implement support for the Animeav1 site, as requested in the issue #3009.

Enables full site support with UI selector for both light and dark themes, and tests for both overview and sync pages.


lolamtisch:
I would recommend to implement the page with chibiscript. Because with chibi the page could be supported directly and you do not need to wait for the next release (probably in a few months).

PR #3076, opened 2025-06-20
Fixes same name item in quicklinks issue

Source-Dom:
Resolve the duplication of quicklinks with the same name by prioritizing chibi quicklinks over page quicklinks.


Source-Dom:
Indirectly fixes #3077

PR #3176, opened 2025-09-12
Update URLs in meta.json for tranimeizle

Musa-Colak:


Source-Dom:
Would it work to put `*` just before the `t` ? @lolamtisch

lolamtisch (changes requested):
Please keep the old domains in case it changes back.

And yeah one can use `*://*.tranimeizle.io/*` to both hit  "*://tranimeizle.io/*" and "*://www.tranimeizle.io/*" in one

Musa-Colak:
I deleted them because the government had blocked them. 😅

lolamtisch (approved):

PR #3180, opened 2025-09-15
add Bato autoUrls

Source-Dom:


henrik9999:
Dont change the entire file formatting

Source-Dom:
> Dont change the entire file formatting

There's no linting at all so I don't really have a choice.

webpackConfig/autoUrls.mjs
```diff
@@ -5,151 +5,181 @@ import path from 'path';
...
-    }
+  let lastError = null;
+  const tasks = {
+    voe,
+    // vidmoly,
+    mixdrop,
+    zoro,
+    // kickassanime,
+    animekai,
+    // bato,
```
lolamtisch: Why did you add bato as comment?

lolamtisch (changes requested):

webpackConfig/autoUrls.mjs
Source-Dom: Forgot to change it back.

lolamtisch (approved):

PR #3190, opened 2025-09-18
Add anizmplayer iframe host for Anizm

Fen1kks:


lolamtisch (changes requested):
`'*://anizmplayer.com/*` is not necessary.

`*://*.anizmplayer.com/*` does map https://ani... too

PR #3196, opened 2025-09-20
Add support to StoneScape

gickass:
Add support to https://stonescape.xyz

### DIFF
--- a/src/pages-chibi/implementations/StoneScape/main.ts
+++ b/src/pages-chibi/implementations/StoneScape/main.ts
@@ -33,6 +33,9 @@ export const StoneScape: PageInterface = {
     getIdentifier($c) {
       return $c.url().urlPart(4).run();
     },
+    getImage($c) {
+      return $c.querySelector('[property="og:image"]').getAttribute('content').ifNotReturn().run();
+    },
     getOverviewUrl($c) {
       return $c
         .string('/series/<identifier>')
@@ -45,6 +48,9 @@ export const StoneScape: PageInterface = {
         .coalesce($c.url().urlPart(5).regex('ch[_-](\\d+)', 1).number().run())
         .ifNotReturn()
         .run();
+    },
+    nextEpUrl($c) {
+      return $c.querySelector('a.next_page').getAttribute('href').ifNotReturn().urlAbsolute().run();
     },
     readerConfig: [
       {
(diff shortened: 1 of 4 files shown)


Source-Dom:
Do `npm run lint:fix` please. Also for future endeavors, you should set your default formatter to `prettier eslint`, if you don't know how, in VS Code, for example, you can right click your file and press `Format document with` and then `Choose default formatter`.

Source-Dom:
Is nextEpUrl supposed to be in manga implementation? Not sure.

gickass:
@Source-Dom I'm not sure too, though I add it because some of manga implementation have it too

src/pages-chibi/implementations/StoneScape/main.ts
```diff
@@ -0,0 +1,129 @@
...
+    getImage($c) {
+      return $c.querySelector('[property="og:image"]').getAttribute('content').ifNotReturn().run();
+    },
+    uiInjection($c) {
+      return $c.querySelector('.summary_image').uiAppend().run();
+    },
+  },
+  list: {
+    elementsSelector($c) {
+      return $c.querySelectorAll('ul li').run();
```
lolamtisch: Too general. It does match more than just the list. You could use something like `.listing-chapters_wrap .wp-manga-chapter`

src/pages-chibi/implementations/StoneScape/main.ts
```diff
@@ -0,0 +1,129 @@
...
+        current: {
+          selector: '.reading-content img',
+          mode: 'countAbove',
+        },
+        total: {
+          selector: '.reading-content img',
+          mode: 'count',
+        },
+      },
+    ],
```
lolamtisch, line 81: The reader can be set to single page mode. This support both. Can you add it and test it?

```
readerConfig: [
      {
        condition: '#single-pager',
        current: {
          selector: '#single-pager [selected="selected"]',
          mode: 'text',
          regex: '^(\\d+)/(\\d+)$',
          group: 1,
        },
        total: {
          selector: '#single-pager option',
          mode: 'text',
          regex: '^(\\d+)/(\\d+)$',
          group: 2,
        },
      },
      {
        current: {
          selector: '.reading-content img',
          mode: 'countAbove',
        },
        total: {
          selector: '.reading-content img',
          mode: 'count',
        },
      },
    ],
```

lolamtisch (changes requested):

src/pages-chibi/implementations/StoneScape/main.ts
gickass: For single page, I think it was the website the problem.
It works if the website load properly after clicking for next page from
https://stonescape.xyz/series/plaything/ch-55/p/1/ -> .../p/2/ .../p/3/ and so on

But the website gave blank if clicking next page from
https://stonescape.xyz/series/plaything/ch-55/?style=paged -> https://stonescape.xyz/series/plaything/ch-55/p/2/
needing to reload the second page to load the website. 
But because it gave blank on next page, the chapter immediately marked as read

PR #3222, opened 2025-09-24
Feat/animecix

Fen1kks:


src/pages/playerUrls.js
```diff
@@ -828,6 +828,10 @@ module.exports = {
   anizmplayer: {
     match: ['*://*.anizmplayer.com/*'],
   },
+  // Animecix
+  tauvideo: {
+    match: ['*://tau-video.xyz/*', '*://*.tau-video.xyz/*'],
```
lolamtisch: I think you know what is wrong

src/pages/list.json
```diff
@@ -1,747 +0,0 @@
-[
```
lolamtisch, line 1: no

src/pages-chibi/implementations/Animecix/main.ts
```diff
@@ -0,0 +1,103 @@
+import { PageInterface } from '../../pageInterface';
```
lolamtisch, line 2: Please check that all the functions work. There enought that do not work like the catalogue handling for example

PR #3238, opened 2025-09-25
[ADD] french-anime.com

FurTorie:
If you manage to patch the episode selector so that it returns the correct data, thank you. But I think the issue comes from the site, which doesn’t have a good system. I wasn’t able to fix it myself, but the rest should be fine.
image


src/pages-chibi/implementations/FrenchAnime/main.ts
```diff
@@ -0,0 +1,74 @@
...
+    elementEp($c) {
+      return $c.text().regex('Episode (\\d+)', 1).number().run();
+    },
+  },
+  lifecycle: {
+    setup($c) {
+      return $c.addStyle(require('./style.less?raw').toString()).run();
+    },
+    ready($c) {
+      return $c.domReady().trigger().run();
```
Source-Dom, line 59:
```suggestion
      return $c.detectURLChanges($c.trigger().run()).domReady().trigger().run();
```

This has to include the detecturlchange function, that's why it doesn't detect when the page's url changes.

Ceci doit inclure la fonction detecturlchange, c'est pour ça que ça ne détecte pas quand l'url de la page change.

src/pages-chibi/implementations/FrenchAnime/main.ts
```diff
@@ -0,0 +1,74 @@
...
+    getIdentifier($c) {
+      return $c.url().urlPart(-1).split('-').first().run();
+    },
+    getImage($c) {
+      return $c.querySelector('.mov-img img#posterimg').getAttribute('src').ifNotReturn().urlAbsolute().run();
+    },
+    getOverviewUrl($c) {
+      return $c.url().run();
+    },
+
```
Source-Dom:
```suggestion
```

src/pages-chibi/implementations/FrenchAnime/main.ts
```diff
@@ -0,0 +1,74 @@
...
+    },
+  },
+  lifecycle: {
+    setup($c) {
+      return $c.addStyle(require('./style.less?raw').toString()).run();
+    },
+    ready($c) {
+      return $c.domReady().trigger().run();
+    },
+
```
Source-Dom:
```suggestion
```

src/pages-chibi/implementations/FrenchAnime/style.less
Source-Dom, line 1: You can't keep it empty usually, because MALSync elements won't look good. Make sure that everything looks good and that everything works properly, including elements that are overlayed, like the progress and when you press C.

Tu ne dois pas le laisser vide d'habitude, car les éléments de MALSync ne seront pas beaux. Assure toi que les éléments sont beau et que tout fonctionne bien, incluant les éléments qui s'affiche par dessus la page, comme le progrès ou que tu appuies sur C.

src/pages-chibi/implementations/FrenchAnime/tests.json
Source-Dom, line 1: For this file, copy from another chibi page folder and change the content to fit this site's data.

Pour ce fichier, copie depuis un autre dossier d'une page chibi et change le contenu par rapport à ce qu'il y a sur ce site.

Source-Dom:
You also have to format your code, use format document in vscode.

Tu dois aussi mettre en forme ton code, utilise "format document" dans vscode.

@FurTorie

FurTorie:
Everything should be fine now. The initial episode detections don’t work, but this is actually better because the site always returns the latest episode. This does not block anime detection: the user just needs to select their episode to trigger the update.
For movies with only one option, they simply need to click "Next".

image

src/pages-chibi/implementations/FrenchAnime/main.ts
```diff
@@ -0,0 +1,80 @@
...
+      // Get the unique post ID with null safety
+      return $c
+        .querySelector('input[name="post_id"]')
+        .ifNotReturn($c.string('no-id').run())
+        .getAttribute('value')
+        .ifNotReturn($c.string('no-value').run())
+        .run();
+    },
+    getOverviewUrl($c) {
+      // Section from URL (variable: exclue, animes-vf, etc.)
```
lolamtisch: `return $c.url().run();` Is probably ok as they all have the same url

src/pages-chibi/implementations/FrenchAnime/main.ts
```diff
@@ -0,0 +1,80 @@
...
+        .querySelector('h1[itemprop="name"]')
+        .text()
+        .trim()
+        .run();
+    },
+    getIdentifier($c) {
+      // Get the unique post ID with null safety
+      return $c
+        .querySelector('input[name="post_id"]')
+        .ifNotReturn($c.string('no-id').run())
```
lolamtisch: Keep the function without content.
`.ifNotReturn()`

src/pages-chibi/implementations/FrenchAnime/main.ts
FurTorie: ❌ return $c.url().run();

Returns the full URL of the current episode
When you switch to a different anime, the entire URL changes
MALSync loses the link between episodes of the same anime
Result: Automatic synchronization no longer works

src/pages-chibi/implementations/FrenchAnime/main.ts
FurTorie: I wasn’t able to get it working properly — it either caused issues or significant slowdowns, and I’m not sure why.

lolamtisch: Without removing it. It is just wrong.

lolamtisch: It is fine in the ready function but not here

src/pages-chibi/implementations/FrenchAnime/main.ts
lolamtisch: The current implementation at least is not correct. Beacuase you can not use ` like that here.
image
You will need to do something similar to this here
https://github.com/MALSync/MALSync/blob/master/src/pages-chibi/implementations/ArvenComics/main.ts#L31

src/pages-chibi/pages.ts
```diff
@@ -29,7 +29,15 @@ import { StoneScape } from './implementations/StoneScape/main';
 import { ResetScans } from './implementations/ResetScans/main';
 import { Utoon } from './implementations/Utoon/main';
 import { WitchScans } from './implementations/WitchScans/main';
+import { Sadscans } from './implementations/Sadscans/main';
```
lolamtisch, line 32: Something did go wrong here. Can you merge or rebase to the current master? It does not compile like this

ttaa-lgtm:
hello new source can be added https://anime-kami.com/

Source-Dom:
> hello new source can be added https://anime-kami.com/

This is unrelated to french-anime ??

FurTorie:
Everything is fine on the tests except for Cloudflare because I don't have an API connection.  
image

src/pages-chibi/implementations/FrenchAnime/main.ts
```diff
@@ -0,0 +1,85 @@
...
+        .run();
+    },
+  },
+  list: {
+    elementsSelector($c) {
+      return $c.querySelectorAll('#epselect option').run();
+    },
+    elementUrl($c) {
+      return $c.url().run();
+    },
```
lolamtisch: You can remove elementUrl completly, as you can not get an url. It will still work then

FurTorie: Done! Removed elementUrl as requested.

PR #3245, opened 2025-09-28
fix title detection for Comikey

gickass:
i hope this is last commit for comikey


gickass:
list still spam in console when clicking homepage from overview/chapter page,
but at least it doesn't spam anymore on chapter page

lolamtisch (approved):

PR #3248, opened 2025-09-28
add support to atsu.moe

gickass:
https://atsu.moe/

### DIFF
--- a/src/pages-chibi/implementations/Atsumaru/main.ts
+++ b/src/pages-chibi/implementations/Atsumaru/main.ts
@@ -1,0 +1,117 @@
+import { PageInterface } from '../../pageInterface';
+
+export const Atsumaru: PageInterface = {
+  name: 'Atsumaru',
+  domain: 'https://atsu.moe/',
+  languages: ['English'],
+  type: 'manga',
+  urls: {
+    match: ['*://atsu.moe/*'],
+  },
+  search: 'https://atsu.moe/search?query={searchtermRaw}',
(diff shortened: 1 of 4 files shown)


Source-Dom:
Closes #3249

src/pages-chibi/implementations/Atsumaru/style.less
```diff
@@ -0,0 +1,19 @@
+@import './../pages';
```
lolamtisch, line 1: I think this is simpler and looks better. But you decide
```less
@import './../pages';

@boxBackground: var(--color-slate3);

@highlightStyle: 1;
@activeEp: var(--color-accent);

#malp {
  margin-top: 15px;
  color: var(--color-foreground2);
}

```

src/pages-chibi/implementations/Atsumaru/main.ts
```diff
@@ -0,0 +1,117 @@
...
+    setup($c) {
+      return $c.addStyle(require('./style.less?raw').toString()).run();
+    },
+    ready($c) {
+      return $c.detectURLChanges($c.trigger().run()).domReady().trigger().run();
+    },
+    listChange($c) {
+      return $c
+        .detectChanges(
+          $c.querySelector('.flex-col > .relative.flex').ifNotReturn().text().run(),
```
lolamtisch: This takes the text of the first list element. This works expect for the first `show all chapters`, as the first element does not change. 
image

PR #3252, opened 2025-09-29
add support to ComicTop

gickass:
https://comic-top.com/

### DIFF
--- a/src/pages-chibi/implementations/ComicTop/main.ts
+++ b/src/pages-chibi/implementations/ComicTop/main.ts
@@ -1,0 +1,168 @@
+import { PageInterface } from '../../pageInterface';
+
+export const ComicTop: PageInterface = {
+  name: 'ComicTop',
+  domain: 'https://comic-top.com/',
+  languages: ['Japanese'],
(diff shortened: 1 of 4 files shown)


src/pages-chibi/implementations/ComicTop/main.ts
```diff
@@ -0,0 +1,160 @@
...
+    },
+  },
+  list: {
+    elementsSelector($c) {
+      return $c.querySelectorAll('.episode-manga li').run();
+    },
+    elementUrl($c) {
+      return $c.find('.epsleft a').getAttribute('href').urlAbsolute().run();
+    },
+    elementEp($c) {
```
lolamtisch, line 142: I think you can just use this. Less duplicate code

`return $c.this('list.elementUrl').this('sync.getEpisode').run();`

lolamtisch (changes requested):

gickass:
i forgot to say. The reason why i remove EpList tests is because url parameter 't' always change so the test always fail.

lolamtisch (approved):

PR #3318, opened 2025-10-09
add support to galaxy degen scans

gickass:
https://gdscans.com/


src/pages-chibi/implementations/GDScans/main.ts
```diff
@@ -0,0 +1,122 @@
...
+    elementEp($c) {
+      return $c.this('list.elementUrl').this('sync.getEpisode').run();
+    },
+  },
+  lifecycle: {
+    setup($c) {
+      return $c.addStyle(require('./style.less?raw').toString()).run();
+    },
+    ready($c) {
+      return $c.detectURLChanges($c.trigger().run()).domReady().trigger().run();
```
lolamtisch: I think `.detectURLChanges($c.trigger().run())` is not necessary here. As the only page that changes the url is the reading page when changing page in the paged view, which triggers a recheck which we do not want.

gickass: i see. I think some of website I did has that problem too. i will fix other later

PR #3322, opened 2025-10-10
fix repeated ready for ResetScans, TrinitiaScans, Utoon

gickass:
Only gdscans.com doesn't need to add much compared to these three website because GDscans would just reload website entirely when changing chapter in page mode.


gickass:
bruh this one is broken because the extension from store is not update yet

PR #3326, opened 2025-10-12
Add Armageddon support

tigeryu8900:
There are two sites: https://armageddontl.com and https://www.silentquill.net. Both sites are owned by KDT Scans; however, since the content of these two sites differs (the former is paywalled and lacks some of the content of the latter), I decided to add them as separate sites.

### DIFF
--- a/src/pages-chibi/implementations/Armageddon/style.less
+++ b/src/pages-chibi/implementations/Armageddon/style.less
@@ -7,11 +7,11 @@
 :root {
   --theme-color: #002e7459;
 
-  &:has(meta[name="theme-color"][content="#ea3364"]) {
+  &:has(meta[name='theme-color'][content='#ea3364']) {
     --theme-color: #ea3364;
   }
 
-  &:has(meta[name="theme-color"][content="#3367d6"]) {
+  &:has(meta[name='theme-color'][content='#3367d6']) {
     --theme-color: #3367d6;
   }
 }


src/pages-chibi/implementations/Armageddon/style.less
```diff
@@ -0,0 +1,21 @@
+@import './../pages';
+
+@boxBackground: var(--box-background);
```
lolamtisch, line 4: The active episode hightlighting does not work on the light mode.

src/pages-chibi/implementations/Armageddon/main.ts
```diff
@@ -0,0 +1,131 @@
...
+      return $c
+        .coalesce(
+          $c.url().urlPart(3).regex('(?:-ch(?:apter)?-(\\d+)(?:-\\d+)?)?$', 1).run(),
+          $c.url().urlParam('chapter').string().split('.').at(0).run(),
+          $c.string('NaN').run(),
+        )
+        .number()
+        .run();
+    },
+    nextEpUrl($c) {
```
lolamtisch, line 44: Breaks when the reader is set to single page.

src/pages-chibi/implementations/Armageddon/style.less
tigeryu8900: I realized the highlight style looks kind of ugly with the site's own hover highlight, so I switched to border style

src/pages-chibi/implementations/Armageddon/main.ts
tigeryu8900: I switched to the chapter drop-down menu since those contain the actual links in both single page and full page

PR #3331, opened 2025-10-13
add support to Mangapill

### LINKED ISSUE #634: Add mangapill.com
Adding this site would be a huge help as this is the site I mainly use for reading manga. The domain is mangapill.com

gickass:
close #634


gickass:
image link blocked by cloudflare, even through inspect element on thumbnail.
I don't know which to use identifier by their numbering(from fourth part of url) or the name, but i choose by the name

PR #3337, opened 2025-10-14
temporary fix for Utoon, ResetScans, TrinitiaScans for webstore extension version

gickass:
Since the .detectChange fix #3257 doesn't apply yet on webstore extension, it make malsync broken on those three website.
so this is only for temporary fix for it.

After malsync webstore version update, probably can revert the code back

### DIFF
--- a/src/pages-chibi/implementations/ResetScans/main.ts
+++ b/src/pages-chibi/implementations/ResetScans/main.ts
@@ -113,7 +113,7 @@ export const ResetScans: PageInterface = {
         .if(
           $c.url().urlPart(5).boolean().run(),
           $c.detectChanges($c.url().urlPart(5).run(), $c.trigger().run()).run(),
-          $c.detectURLChanges($c.trigger().run()).run(),
+          $c.run(),
         )
         .domReady()
         .trigger()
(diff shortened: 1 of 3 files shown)


src/pages-chibi/implementations/ResetScans/main.ts
```diff
@@ -110,7 +110,11 @@ export const ResetScans: PageInterface = {
     },
     ready($c) {
       return $c
```
lolamtisch, line 112: Would it not possible to only use domReady? As far as I can see do the pages (Just looked at reset) not have js based navigation (not a single page application (SPA)). Meaning it always loads a new page and dectectURLChanges is not necessary. 
```
ready($c) {
      return $c
        .domReady()
        .trigger()
        .run();
}
```

gickass: Yeah, detectURLChanges is not necessary but...

the If function is needed for paged style or single page mode on those 3 link.
detectChanges on url part 5 is needed for single page mode because the page doesn't reset when changing chapter or clicking for next/previous chapter in paged style mode.

I think they use SPA on paged style mode right? It great that it doesn't reset page when changing page, but it also doesn't reset when changing chapter too

lolamtisch (approved):

Issue #2875, opened 2025-03-08
Adding a sorting option by latest updated by episode or chapter


TamamoTomato:
### Suggestion

Bump on the title. Hard to keep track of your follows on the latest chapter/episode added.

PR #3339, opened 2025-10-15
Fix and extend minimal sorting

lolamtisch:
- Handle anime entries with last timestamp in the prediction sorting too
- Fix seemingly random entries in the unread sorting
- #2875


Source-Dom:
@lolamtisch Sorry for long response time, this update works really good, thank you so much ! 

[_For this image one entry didn't get ordered, but then I removed the rating and added it back and it fixed itself, so I probably had to clear cache ?_]
image

PR #3342, opened 2025-10-16
Add support to mangaball

### LINKED ISSUE #3274: [Website Request] Mangaball
### Website name

The name of the website is Mangaball

### Website link

https://mangaball.net/

### Website language(s)

The main website is in english, but the titles are also in other languages (alternatives names) and there's chapters in multiple languages such as French, Arabic, Catalan, Chinese, Japanese and much more 

### Other details

I know this is a new website but it's promising and in my opinion it's one with good ui and tracking chapters on it would be awesome. Also to take notes of, there's also NSFW stuff on the website (but you can partially or totally hide it)

gickass:
close #3274
https://mangaball.net/

I don't know if that is the best way of finding overviewUrl in the chapter page

### DIFF
--- a/src/pages-chibi/implementations/MangaBall/style.less
+++ b/src/pages-chibi/implementations/MangaBall/style.less
@@ -1,8 +1,9 @@
 @import './../pages';
-@activeEp: inherit;
 @boxBackground: #0000006e;
 
-body[class^='theme-'] .bg-dark.mal-sync-active {
-  background-color: #ff67401a !important;
-  border: 1px solid #ff67404d !important;
+@highlightStyle: 0;
+@activeEp: var(--accent);
+
+.mal-sync-active .chapter-translation {
+  border-left: 5px solid @activeEp;
 }


src/pages-chibi/implementations/MangaBall/main.ts
```diff
@@ -0,0 +1,124 @@
...
+      return $c.this('sync.getOverviewUrl').this('overview.getIdentifier').run();
+    },
+    getImage($c) {
+      return $c.querySelector('[property="og:image"]').getAttribute('content').ifNotReturn().run();
+    },
+    getOverviewUrl($c) {
+      return $c
+        .querySelector('.yoast-schema-graph')
+        .html()
+        .regex('url"\\s*:\\s*"[^"]*(\\/title-detail\\/[^"]+)', 1)
```
lolamtisch, line 28: You could use json parse and then filter thru the objects. But it probably is fine as is.

src/pages-chibi/implementations/MangaBall/style.less
```diff
@@ -0,0 +1,8 @@
+@import './../pages';
```
lolamtisch, line 1: I think this would be better as it follows the theme colors too. Or what do you think?
```less
@import './../pages';
@boxBackground: #0000006e;

@highlightStyle: 0;
@activeEp: var(--accent);

.mal-sync-active .chapter-translation {
  border-left: 5px solid @activeEp;
}

```

gickass: i think this is better since it look simple and less messy than beforee

src/pages-chibi/implementations/MangaBall/main.ts
gickass: ohh, i just test it but it cannot in this case right? because, the html also contain
```
 <b>Warning</b>:  Undefined array key "chapter_name" in <b>/www/wwwroot/mangadex/views/frontend/pages/chapter/chapterIndex.php</b> on line <b>71</b><br />
" }
```

PR #3346, opened 2025-10-18
Convert full width character to half width

gickass:
for some reason, ChibiGeneratorTypes.ts cannot commit with it. this is what i wrote:
  ```
/**
   * Converts input to half-width character
   * @input string
   * @param value - Optional static string value to return instead of converting input
   * @returns A string representation of the input or the static value
   * @example
   * $c.string('Ｈｅｌｌｏ Ｗｏｒｌｄ ３').convertCh().run(); // returns Hello World 3
   * @see {@link  functionsRegistry.convertCh} for implementation details
   */
  convertCh: CT.MatchesType<Input, CT.InputType<typeof functionsRegistry.convertCh>> extends true
    ? (value?: string) => ChibiGenerator<ReturnType<typeof functionsRegistry.convertCh>>
    : CT.TypeMismatchError<'convertCh', Input, CT.InputType<typeof functionsRegistry.convertCh>>;
```

I do this by copy and paste part of other code like .number() and .string(). I test it and it should work

### DIFF
--- a/src/chibiScript/functions/literalFunctions.ts
+++ b/src/chibiScript/functions/literalFunctions.ts
@@ -53,4 +53,19 @@ export default {
   object: (ctx: ChibiCtx, input: void, value: object) => {
     return value;
   },
+
+  /**
+   * Converts input to half-width character
+   * @input any
+   * @param value - Optional static string value to return instead of converting input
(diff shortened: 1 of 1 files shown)


gickass:
wrong branch from

PR #3347, opened 2025-10-18
add function to converts characters to half-width character

gickass:
i don't know naming it as .convertCh is suitable. Can use a better name to make it more recognizeable


lolamtisch (changes requested):
Could you add a testcase too?

https://github.com/MALSync/MALSync/blob/chibiDev/test/src/chibiScript/functions/stringFunctions.test.ts

lolamtisch:
Sorry forgot about the name. Yeah convertCh is a bit to generic. Something like `toHalfWidth` or `normalizeWidth` is probably better.

PR #3351, opened 2025-10-19
Convert animepahe to chibi

tigeryu8900:
Converts animepahe to chibi, to be merged after next release

### DIFF
--- a/src/pages-chibi/implementations/animepahe/main.ts
+++ b/src/pages-chibi/implementations/animepahe/main.ts
@@ -6,6 +6,7 @@ export const animepahe: PageInterface = {
   languages: ['English'],
   type: 'anime',
   database: 'animepahe',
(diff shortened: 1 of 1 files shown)


lolamtisch:
I am not sure if settings the [minimumVersion](https://github.com/MALSync/MALSync/blob/master/src/pages-chibi/pageInterface.ts#L51) will stop the chibi version to override the old implementation.


If it still overrides you could do this as an PR to the [chibiDev](https://github.com/MALSync/MALSync/tree/chibiDev) branch. I will merge it before the next release and then stop chibi builds until it is live in the stores.

tigeryu8900:
Just confirmed, setting `minimumVersion` to `0.12.1` does prevent the chibi version from overriding the old version.

lolamtisch (approved):

PR #3367, opened 2025-10-27
Armageddon add theme color #fc7500

tigeryu8900:
Ideally, we should be able to set the color dynamically by the meta element `meta[name='theme-color']`, but this is just a quick fix.


tigeryu8900:
I merged the changes into #3589

Issue #3369, opened 2025-10-27
add `user-select: none;` to malsync-rel-link to stop it from being copied from title.


Source-Dom:
> ok I've found a solution we need to add `user-select: none;` to malsync-rel-link.

 _Originally posted by @Source-Dom in [#3368](https://github.com/MALSync/MALSync/issues/3368#issuecomment-3449861605)_

Issue #3368, opened 2025-10-27
MAL button on the title placement


Gusatvoschaivon:
I'm having a problem when i sometimes just copy past the title of a title, it comes with a #  and the MAL link with an image lol

<img width="928" height="117" alt="Image" src="https://github.com/user-attachments/assets/1d053699-bc4d-48a6-b738-e6811d8e4e4c" />

In some places it just goes like:
"  One Piece

"
With a space behind and a line break after, a bit annoying. 

The header thing I think is an anilist problem, but the MAL link could be in an another layer or something, like the anilist button is on MAL

<img width="1473" height="279" alt="Image" src="https://github.com/user-attachments/assets/1c69b202-e367-4f8f-947d-095292f2ad6d" />

Source-Dom:
Can you give some links, I don't understand where that is.

Gusatvoschaivon:
On any anime/manga page, Right by the title. Maybe it's a opinion that make that appear for me? If so I think it was on by default

Gusatvoschaivon:
I'm not on the computer now, but it's right here 

![Screenshot_20251027_025045_Firefox.jpg](https://github.com/user-attachments/assets/023c4675-79c1-4aad-a25c-8af54fdde4c3)

Source-Dom:
this ? the mal logo ?

<img width="1405" height="440" alt="Image" src="https://github.com/user-attachments/assets/d3e96dc9-6226-4862-8adc-a02412a13767" />

Gusatvoschaivon:
Yeah, yours is also connected to the title on the HTML?

Source-Dom:
Yeah. I don't understand how you got nothing more than the title when you copy it tho.

Like I copy the whole page and don't even get the myanimelist link.

Can you say step by step exactly what you're doing, so I can reproduce.

Gusatvoschaivon:
Sometimes I double click and other times I tried to select the words, but if I go a bit to the side it picks ups the mal icon

Also I'm on Firefox, using the extension and one of the places that pickup the icon is obsidian, but as I said when I try on someplaces it gives me a line break after the title

Source-Dom:
So when you paste on obsidian it picks up the logo link ?

Source-Dom:
use paste as plain text or CTRL + SHIFT + V (shortcut)
<img width="259" height="406" alt="Image" src="https://github.com/user-attachments/assets/091853f8-2af7-4937-9412-c94bbd9f9e2d" />

[...]

PR #3371, opened 2025-10-27
fix malsync-rel-link from being able to be copied in kitsu

Source-Dom:
add `user-select: none` to kitsu in malsync-rel-link

#3368 #3369


Source-Dom:
Sorry for the 2 prs I was too lazy to open vscode <3 #3370

PR #3375, opened 2025-10-28
fix #3373 - "close" button getting hidden on small devices

### LINKED ISSUE #3373: Imposible to Close Correction Box on Mobile Landscape
### Describe the Bug

It is impossible to close the Correction popup when on mobile in landscape. The close Text is not visible and the correction app does not have internal scrolling.

The workaround is to put the screen back into portrait and then close there. A fix is to add an X in the `m-s-pill-section`.

1. have screen in landscape
2. Browse to a anime site, say animekai.to
3. Select an episode
4. Select the pencil in the bottom right thats in the circle
5. The correction screen pops up with "Settings", "URL", "Correction Offset", etc
6. the close button at the bottom of the box is not visible and it is impossible to scroll down

### Browser

Firefox Android Beta 144(?)

### Type

Extension

### URL

https://animekai.to/watch/a-journey-through-another-world-raising-kids-while-adventuring-ll3k#ep=5

Source-Dom:
Makes it so close button never gets hidden based on screen height, in correction menu popup.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,20 +3,9 @@
   // https://marketplace.visualstudio.com/items?itemName=stylelint.vscode-stylelint#optional-but-recommended-setup
   "css.validate": false,
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "editor.rulers": [
-    120
-  ],
-  "eslint.validate": [
-    "javascript",
-    "typescript",
-    "vue"
-  ],
-  "stylelint.validate": [
-    "css",
-    "scss",
-    "vue",
-    "less"
-  ],
+  "editor.rulers": [120],
+  "eslint.validate": ["javascript", "typescript", "vue"],
+  "stylelint.validate": ["css", "scss", "vue", "less"],
   "less.validate": false,
   "scss.validate": false,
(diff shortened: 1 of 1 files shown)


src/_provider/Search/correctionStyle.less
```diff
@@ -5,22 +5,31 @@
...
 
   .close {
     display: block;
-    margin-top: 15px;
+    line-height: 36px;
     text-align: center;
+    position: sticky; /* keep visible at bottom when content scrolls */
     cursor: pointer;
+    bottom: 0;
+    background: #323232;
```
lolamtisch, line 32: Why the backgroundcolor?

Source-Dom: It makes it so the close button has something behind or it will appear alone with a transparent background and will be hard to read.

Source-Dom: Also it doesn't hide anything behind it, because when you scroll to the bottom the last text field will appear on top of the button not under.

Source-Dom:
fix #3373

lolamtisch (approved):

PR #3378, opened 2025-10-29
.waitUntilTrue test at chibiDev

gickass:
for comicdays, and test for .waitUntilTrue in OverviewIsReady on stonescape.xyz for chibiDev


gickass:
I test it by using the query that is not exist, to check that if it available, but, the query it still go through overviewIsReady/syncisready as if that query available.

It should not pass the test, but this pr still let tests pass like it skipped .waitUntilTrue command in ChibiDev

PR #3379, opened 2025-10-29
waitUntilTrue-test for master branch

gickass:
This is test at master branch for stonescape.xyz that use waitUntilTrue in OverviewisReady


gickass:
same with other pr, I test it by using the query that is not exist, to check that if it available, and it gave timeout because the query will never exist to return true.

It should not pass the test, so this show .waitUntilTrue work perfectly in master

PR #3386, opened 2025-11-02
add Ritharscans support

tigeryu8900:


tigeryu8900:
The `0.1` in `test/dist/test.js` results in the error `Invalid Version: 0.1`, so I changed it to `0.1.0`.

src/pages-chibi/implementations/Ritharscans/main.ts
```diff
@@ -0,0 +1,126 @@
...
+    getEpisode($c) {
+      return $c
+        .getGlobalVariable('data')
+        .get('name')
+        .string()
+        .regex('^Chapter (\\d+)', 1)
+        .number()
+        .run();
+    },
+    nextEpUrl($c) {
```
lolamtisch, line 41: Does not seem to work. Probably better to directly get it with regex

tigeryu8900: Fixed. Seems like it's not proper json, so it fails to parse.

PR #3394, opened 2025-11-05
add function to convert japanese numeral to standard numeral

gickass:


gickass:
i notice from first test, this return as number instead of string. Should I leave it or turn it into string?

Source-Dom:
`getEpisode` and `getVolume` require a number type, so you can leave as number I believe -> 
https://github.com/MALSync/MALSync/blob/1307037d76f6f32f7a658e39e260770bb4c0eb06/src/pages-chibi/pageInterface.ts#L94

https://github.com/MALSync/MALSync/blob/1307037d76f6f32f7a658e39e260770bb4c0eb06/src/pages-chibi/pageInterface.ts#L101

test/src/chibiScript/functions/stringFunctions.test.ts
```diff
@@ -211,6 +211,33 @@ describe('String Functions', () => {
...
+      const code = $c.string('二百三').JPtoNumeral().run();
+      expect(generateAndExecute(code).run()).to.equal(203);
+    });
+
+    it('should converts Japanese thousand to standard Arabic numeral', () => {
+      const code = $c.string('二千八百七十一').JPtoNumeral().run();
+      expect(generateAndExecute(code).run()).to.equal(2871);
+    });
+
+    it('should converts Japanese ten thousand numeral to standard Arabic numeral', () => {
```
Source-Dom: Again

PR #3406, opened 2025-11-12
add getMalUrl and :hover style in atsumaru

gickass:


src/pages-chibi/implementations/Atsumaru/main.ts
```diff
@@ -83,6 +83,25 @@ export const Atsumaru: PageInterface = {
...
+
+      const getAnilist = $c
+        .provider()
+        .equals('ANILIST')
+        .ifNotReturn()
+        .querySelector('.aspect-square[title="AniList"]')
+        .getAttribute('href')
+        .run();
+
+      return $c.coalesce($c.fn(getMal).run(), $c.fn(getAnilist).run()).ifNotReturn().run();
```
Source-Dom:
```suggestion
      return $c.coalesce(getMal, getAnilist).ifNotReturn().run();
```

Try this, it should work. Unless it doesn't ? Previous implementation makes you run the code twice, technically.

gickass: hmm 
onsite replace
it only work for first coalesce function when i use myanimelist in testing, the second getAnilist doesn't run when I use anilist when testing if i remove fn

Source-Dom: > it only work for first coalesce function, the second getAnilist doesn't run when I set the provider as anilist if i remove fn

are you using myanimelist as your main list, when  testing ?

gickass: i'm using both myanimelist and anilist when testing
edit: i mean by using myanimelist in other test, and anilist in other test

Source-Dom: you can't return all of those at once, it doesn't make sense the implementation doesn't handle multiple string, only one.

Source-Dom: what do the logs look like when you're running your implementation with fn ?

gickass: > you can't return all of those at once, it doesn't make sense the implementation doesn't handle multiple string, only one.

i mean, i'm using 1 provider per testing. when using myanimelist, it do return myanimelist when it show Overwrite in onsite url in console.

But when i use anilist as provider when testing, it doesn't return anything when i remove fn()
I tried doing similar getMalUrl from Comix implementation though

gickass: > what do the logs look like when you're running your implementation with fn ?

it shows this when using anilist
when using fn

show this when using myanimelist
image

show this error when there are no myanimelist/anilist in the page depend on what provider i'm using
when error

Source-Dom: oh you're right that's so weird

gickass: hmm maybe because of .coalesce that kinda have it to technically run twice

Source-Dom: I guess you can keep it like that I can't make it work with whatever I try :/

lolamtisch: All provider can handle mal urls. Meaning mal can always be returned. Default in the past was if mal url exist return that if not then if provider anilist then that if not nothing. Later on we deciced that it makes sense to always return anilist url if it exists and then fallback to mal.

Maybe we should think about removing getMalUrl form chibi and replace it with something better and handle the logic outside of chibi.

Something like

```ts
Provider: { // no idea what to call the section
    getMal: ($c) => $c.querySelector("a[href^='https://myanimelist.net']").getAttribute('href').run(),
    getAnilist: ($c) => $c.querySelector("a[href^='https://anilist.net']").getAttribute('href').run(),
}
```

PR #3429, opened 2025-11-20
Add new match URL pattern for aniwatchtv.to

Jayanth-9090:
Added the domain aniwatchtv.to to Zoro/Hianime since they use the same url mappings


henrik9999:
This is a fake site and doesn't have to do anything with the real site

PR #3430, opened 2025-11-21
Add support to comic walker

gickass:
https://comic-walker.com/


gickass:
i might miss some regex at episode detection

PR #3436, opened 2025-11-22
Add support to LuaScans

gickass:
https://luacomic.org/


src/pages-chibi/implementations/LuaScans/style.less
```diff
@@ -0,0 +1,20 @@
+@import './../pages';
+
+@boxBackground: #262626;
```
lolamtisch: Does break on light theme.

Did not test it but using `hsl(var(--secondary))` should fix this.

PR #3464, opened 2025-12-01
Mangabaka support

lolamtisch:


src/_provider/MangaBaka/helper.ts
```diff
@@ -0,0 +1,179 @@
...
+        throw new NotAutenticatedError('user_token_failed');
+      default:
+        throw new Error(
+          res.error_description ? `${res.error_description} (${res.error})` : res.error,
+        );
+    }
+  }
+}
+
+export function bakaStateToState(input: BakaState): status {
```
jippi, line 160: `considering` has been added that in most cases map to Plan To Read in MAL/AL

src/_provider/MangaBaka/list.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  authenticationUrl = authenticationUrl;
+
+  async getUserObject() {
+    const json = (await call(urls.userInfo())) as { name: string; preferred_username: string };
+    console.log(json);
+    return {
+      username: json.name,
+      picture: 'https://mangabaka.org/images/logo.png',
+      href: `https://mangabaka.org/u/${json.preferred_username}`,
```
jippi: usernames are optional, so might be empty - unsure if it matters here or not :)

src/_provider/MangaBaka/metaOverview.ts
```diff
@@ -0,0 +1,366 @@
...
+        gen.push({
+          text: i.name,
+          url: `https://mangabaka.org/search?genre=${i.name}`,
+        });
+      });
+      this.meta.info.push({
+        title: api.storage.lang('overview_sidebar_Genres'),
+        body: gen,
+      });
+    }
```
jippi, line 225: this is likely to change before v2 switch over to be the primary, the URL part at least - it should use the ID at least in the URL instead of name

src/_provider/MangaBaka/search.ts
```diff
@@ -0,0 +1,37 @@
...
+      url: `https://mangabaka.org/${item.id}`,
+      malUrl: () => {
+        return Promise.resolve(
+          item.source.my_anime_list?.id
+            ? `https://myanimelist.net/manga/${item.source.my_anime_list.id}`
+            : null,
+        );
+      },
+      image: item.cover.x150.x2 || '',
+      imageLarge: item.cover.x350.x2 || '',
```
jippi: unsure if your browser support will work for it, but you can append `.avif` or `.webp` to the URL to get those image formats

we prefer `avif` on our website, so those are likely pre-cached and thus would load faster

src/_provider/MangaBaka/single.ts
```diff
@@ -0,0 +1,306 @@
...
+
+  authenticationUrl = authenticationUrl;
+
+  protected handleUrl(url) {
+    if (url.match(/mangabaka\.(dev|org)\/\d*(\/|$)/i)) {
+      this.type = 'manga';
+      this.ids.baka = Number(utils.urlPart(url, 3));
+      return;
+    }
+    if (url.match(/anilist\.co\/(anime|manga)\/\d*/i)) {
```
jippi, line 41: Could probably do this instead and avoid the type check later in the `if`?

```suggestion
    if (url.match(/anilist\.co\/manga\/\d*/i)) {
```

src/_provider/MangaBaka/single.ts
```diff
@@ -0,0 +1,306 @@
...
+  _getTotalVolumes() {
+    return Number(this.libraryEntry.Series.final_volume) || 0;
+  }
+
+  _getDisplayUrl() {//TODO: Check
+    return this.displayUrl !== '' && this.displayUrl !== null ? this.displayUrl : this.url;
+  }
+
+  _getImage() {
+    return this.libraryEntry.Series.cover.x350.x2 || '';
```
jippi: same here re `.avif` and `.webp` suffix

src/_provider/MangaBaka/types.ts
```diff
@@ -0,0 +1,230 @@
+export type BakaState =
```
jippi, line 1: `considering` is also a state

src/_provider/MangaBaka/helper.ts
```diff
@@ -0,0 +1,198 @@
...
+      return res;
+    });
+}
+
+export function errorHandling(res, code) {
+  if ((code > 499 && code < 600) || code === 0) {
+    throw new ServerOfflineError(`Server Offline status: ${code}`);
+  }
+
+  if (res && (res.status < 200 || res.status >= 300)) {
```
jippi, line 130: do you handle 429 in other places for rate limited?

src/_provider/MangaBaka/helper.ts
```diff
@@ -0,0 +1,198 @@
...
+        throw new Error(
+          res.message
+            ? `${res.message} (${res.status})`
+            : `Unknown Error, status code: ${res.status}`,
+        );
+    }
+  }
+
+  // Userinfo call
+  if (res && typeof res.error !== 'undefined') {
```
jippi, line 146: any bad/invalid/missing auth error should come with 401 response code, so wouldn't the switch above handle that?

src/_provider/MangaBaka/helper.ts
```diff
@@ -0,0 +1,198 @@
...
+      return 'considering';
+    case status.All:
+      return null;
+    default:
+      throw new Error(`Unhandled Status: ${input}`);
+  }
+}
+
+export function getAlternativeTitles(series: BakaSeries): string[] {
+  const titles: string[] = [];
```
jippi, line 205: there are cases where `title` might be identical to `romanized_title` or `native_title` - so if you plan to display them, I would check if the value is in `titles` before pushing them

src/_provider/MangaBaka/list.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  private limit = 100;
+
+  async getPart() {
+    if (this.listType !== 'manga') {
+      throw new Error('MangaBaka only supports manga');
+    }
+
+    this.limit = 100;
+    if (this.offset < 1) this.offset = 1;
```
jippi: Can also do 
```suggestion
   this.offset = Math.max(1, this.offset)
```

src/_provider/MangaBaka/helper.ts
lolamtisch: Yeah it is handled centralised. With a bit of crude retry logic

https://github.com/MALSync/MALSync/blob/master/src/background/messageHandler.ts#L205

src/_provider/MangaBaka/helper.ts
lolamtisch: I think most places where that is important does deduplication.

src/_provider/MangaBaka/types.ts
```diff
@@ -228,3 +228,13 @@ export interface SearchResponse {
...
   data: BakaSeries[];
 }
+
+export type ElementReadyEvent = Event & {
+  detail: {
+    element_id: string;
+    name: 'after-links';
+    series: BakaSeries;
+    list_config: any;
+    user: any;
```
jippi, line 239: At time of writing, this is how these types are defined - subject to change at any time - but might be useful 

```ts
export const OauthScope = z.enum([
	// My Library
	'library.read',
	'library.write',
	// MangaBaka moderator role
	'mod',
	// Basic (required) OIDC scopes
	'openid',
	'profile',
])

export const AuthIdentify = z.object({
	// Required fields
	id: z.string(),

	// "Profile" fields
	nickname: z.string().trim().nullish(),
	preferred_username: z.string().trim().nullish(),
	updated_at: z.coerce.date(),
	role: z.enum(['user', 'developer', 'moderator', 'admin']),

	// Informational
	auth_type: AuthMethod,
	scopes: z.array(OauthScope),
	client: z
		.object({
			name: z.string().trim(),
			icon: z.string().trim().nullable(),
			type: z.enum(['web', 'native', 'user-agent-based', 'public']),
		})
		.nullable(),
})
export const ListConfiguration = z.object({
	mode: ListMode,
	title_romanized: BooleanDefaultFalse,
	title_native: BooleanDefaultFalse,
	publishers: BooleanDefaultTrue,
	staff: BooleanDefaultTrue,
	anime: BooleanDefaultFalse,
	description: BooleanDefaultFalse,

	show_headline: BooleanDefaultTrue,
	multi_column: BooleanDefaultTrue,
	max_picture_dpi: z.coerce.number().min(1).max(3).default(2),

	mb_meta: BooleanDefaultTrue,
	source_al: BooleanDefaultTrue,
	source_ann: BooleanDefaultTrue,
	source_ap: BooleanDefaultTrue,
	source_kt: BooleanDefaultTrue,
	source_mal: BooleanDefaultTrue,
	source_md: BooleanDefaultTrue,
	source_mu: BooleanDefaultTrue,
	source_shikimori: BooleanDefaultTrue,

	// --------------------
	// internal
	// --------------------

	// Apply a line clamp of 1 to title or
	title_only_one_line: BooleanDefaultFalse,

	// Show or hide all rating options
	show_ratings: BooleanDefaultTrue,

	// Hide all titles
	show_titles: BooleanDefaultTrue,

	// Hide all meta (except for volume + chapter)
	show_meta: BooleanDefaultTrue,
})
```

jippi:
@lolamtisch from Discord (https://discord.com/channels/@me/1358450902465511514/1456396609154978004) to implement

lolamtisch:
https://discord.com/channels/1345354242163998764/1345354242709520466/1456622487231271005

jippi:
https://discord.com/channels/@me/1358450902465511514/1457173379282112606

PR #3494, opened 2025-12-16
Make text in anime title easily selectable

Source-Dom:
The title (in media-link) that can be clicked to redirect to provider, is hard to select directly, because it's draggable because of the a tag, so I added `draggable=false` to it, so you can easily select the text.
<br />
Recording 2025-12-16 at 13 47 31

### DIFF
--- a/src/_minimal/views/overview.vue
+++ b/src/_minimal/views/overview.vue
@@ -323,6 +323,7 @@ const totalLoading = computed(() => {
     .header-title {
       position: relative;
       width: 100%;
+      cursor: text;
 
       .header-synonyms {
         display: none;


lolamtisch:
A link is a link, so cursor text does not make sense. Additionally does it not work on chrome. It is not worth the effort really

Source-Dom:
It says draggable attribute is supported on both browsers, I could remove the cursor tho.

lolamtisch:
Yeah but chrome directly opens the link. So it is not possible to select the text. 

I mean one can always use the alt key to select the text. So i really do not understand the usecase other than on mobile and I dont think draggable is a solution there either.

Source-Dom:
🤷 idk I tried the only solution I found, for that person that complained that it wasn't easy to select the text because there's a button in the span. Idk what else can be done to make it easy, but I also don't understand what's the point of making things draggable it's not really useful.

PR #3536, opened 2026-01-08
add utility function slugify

gickass:
slugify title to replace space with hyphen, remove or replace special character to make it suitable as identifier.

Probably for like website that they didn't have anything that can be use for unique identifier for title, like tappytoon

### DIFF
--- a/src/chibiScript/utilities/stringUtilities.ts
+++ b/src/chibiScript/utilities/stringUtilities.ts
@@ -12,10 +12,11 @@ export default {
 
   /**
    * Slugify a string (Converts string to lowercase, removes special characters, and replaces spaces with hyphens)
-   * @param replacement Additional replacement rules in the format {'target': 'replacement'}.
+   * @param replacement Additional replacement rules in the format {'target': 'replacement'} or [['target', 'replacement']].
    * @example
-   * $c.string('1 World/s New').slugify().run() // return 1-worlds-strong
-   * $c.string('100% World-&-☢').slugify([['%', 'percent'], ['☢', $c.string('radiation').run()], ['&', 'and']]).run() // return 100percent-world-and-radiation
+   * $c.string('1 World/s New').slugify().run() // return 1-worlds-new
+   * $c.string('100%-&').slugify([['%', 'percent'], [$c.string('&').run(), $c.string('and').run()]]).run() // return 100percent-and
+   * $c.string('100%-&').slugify({'%': 'percent', '&': $c.string('and').run()}).run() // return 100percent-and
    */
   slugify: (
     $c: ChibiGenerator<string>,


src/chibiScript/utilities/stringUtilities.ts
```diff
@@ -9,4 +9,26 @@ export default {
   replaceLinebreaks: ($c: ChibiGenerator<string>, replacement: ChibiParam<string> = ' ') => {
     return $c.replaceAll('(\n|\r)', replacement);
   },
+
+  /**
+   * Slugify a string (Converts string to lowercase, removes special characters, and replaces spaces with hyphens)
+   * @param values Additional replacement rules in the format 'target:replacement'.
+   * @example
+   * $c.string('1 World/s New').slugify().run() // return 1-worlds-strong
+   * $c.string('100% World-&-☢').slugify('%:percent', '☢:radiation', '&:and').run() // return 100percent-world-and-radiation
```
lolamtisch: I am not sure, but would using a object not make it easier to work with?

```ts
slugify({
  '%': 'percent', 
  '☢': 'radiation',
  '&': 'and',
  ':': 'or'
})
```

gickass: defnitely better since it can use ':' for word replacement too and flexible to use chibiParam on the object value.

I tried to make it able to use chibiParam on object key too, but it need to use array formatting instead 
which make it a bit complicated like double square bracket at the start and at the end,
though it also unnecessary probably to use chibiParam in object key

src/chibiScript/utilities/stringUtilities.ts
```diff
@@ -9,4 +9,24 @@ export default {
...
   },
+
+  /**
+   * Slugify a string (Converts string to lowercase, removes special characters, and replaces spaces with hyphens)
+   * @param replacement Additional replacement rules in the format {'target': 'replacement'}.
+   * @example
+   * $c.string('1 World/s New').slugify().run() // return 1-worlds-strong
+   * $c.string('100% World-&-☢').slugify({'%': 'percent', '☢': $c.string('radiation').run(), '&': 'and'}).run() // return 100percent-world-and-radiation
+   */
+  slugify: ($c: ChibiGenerator<string>, replacement?: Record<string, ChibiParam<string>>) => {
```
lolamtisch: Not sure but could the key theoretically not be a ChibiParam too?

gickass: I'm actually not sure too... but it probably not...?

chibiparam try
I tried and change the Object key type to ChibiParam<string>, and it only gave error at compile:

> Type 'ChibiParam<string>' does not satisfy the constrain 'string | number | symbol'.

Then I try to test with key after that compile with:
`$c.string('1 World / s\n-\n\n New').slugify({[$c.string('/').run()]: 'or'}).run();`
it will ignore the key value (which is '/'), so it will remove '/' instead because it doesn't replace and regex later remove it. 

It also gave warning:

> A computed property name must be of type 'string', 'number', 'symbol', or 'any'.

image

Also  writing it like this:
```
 slugify: (
    $c: ChibiGenerator<string>,
    replacement?: { [key: ChibiParam<string>]: ChibiParam<string> },
  ) => {
    const expr = Object.entries(replacement ?? {}).reduce(
      (acc, [from, to]) => acc.replaceAll(from, to),
      $c.toLowerCase(),
    );

    return expr
      .replaceRegex('[^\\w\\s-]', '')
      .trim()
      .replaceRegex('\\s+', '-')
      .replaceRegex('-{2,}', '-');
  },
```
 
gave same as above error

if I need to make key as chibiParam too, it probably have to use array instead of object;
so it will write a function in array to use it... but it kinda make it a bit more complicated to use it.
```
slugify([
  [$c.string('%').run(), $c.string('percent').run()],
  [$c.string('☢').run(), $c.string('radiation').run()],
])
```

PR #3551, opened 2026-01-15
Implement Matrix for AutoUrls (parallelism)

Source-Dom:
- Launches all pages parallelly with Matrix, which runs separate jobs for each page, at once.
- Use upload/download-artifact to build an initial reusable state to not do `npm install` in each separate jobs, for efficiency.

<br>

- Add better logging. Show found pages/chibi/players per site.
- Show final result by showing if tasks is succesful or failed.
- Add custom error annotations.


Source-Dom:
it seems my master branch didn't want to sync, I'll fix it when I can, so I didn't have bato and the other one, that's why it's deleted

Source-Dom:
@lolamtisch merge ready ?

PR #3610, opened 2026-02-09
feat: fireani.me support

senpaizeroani-design:
This is my attempt implementing fireani.me 
The handling is kinda shakey as the website is an spa but tests and inside chrome it seems to work. Data is being parsed from special #malsync element provided by the frontend for easy implementation.
I could also expose the anilist_id but it seems there is only a function getMalUrl so i'm not sure if this could be added.

### DIFF
--- a/src/pages-chibi/implementations/FireAnime/main.ts
+++ b/src/pages-chibi/implementations/FireAnime/main.ts
@@ -18,7 +18,8 @@ export const FireAnime: PageInterface = {
       return getJsonData($c).get('title').run();
     },
     getIdentifier($c) {
-      return getJsonData($c).get('slug').run();
+      // To handle every season separately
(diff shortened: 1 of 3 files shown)


henrik9999:
We have a new way to add pages described in the wiki the ones in the pages folder are old implementations

henrik9999:
https://github.com/MALSync/MALSync/blob/master/src/pages-chibi/implementations/AnimeKAI/main.ts

This one could help you as it also uses json, there you can also see that getmalurl supports anilist but if malid exists that should be returned first

senpaizeroani-design:
@henrik9999  gut stuck with the chibi script, I suspect it fails during the parsing of the json of the size, thow not sure how to catch that in chibi.
any help would be appreciated.

senpaizeroani-design:
lgtm

PR #3623, opened 2026-02-13
add support to allManga

### LINKED ISSUE #2574: [Website Request] Allmanga.to
### Website name

AllManga

### Website link

https://allmanga.to/

### Website language(s)

English

### Other details

Even though the name is all manga it is also used anime and the anime quality and library is good so i would appreciate if you put this as an anime website.

gickass:
close #2574 
this site use ads url even after using ublock
image

when compute type on manga, it cause readerconfig not applied on manga page causing immediate sync. 
If reversed (where base type is manga compute type is anime), cause immediate sync on anime,

i think it better to only have it immediate sync on manga, rather than of anime so setting the base type as anime is better


lolamtisch:
application/ld+json is empty for me

image

gickass:
i cannot make malsync detect the player for those 2 new link i add for unknown reason.

PR #3684, opened 2026-03-07
Reduce AniList calls

BrutuZ:
Replace `Pages` (limited to 50 entries per call) with `MediaListCollection` (up to 500 entries per call) on the GraphQL query, drastically reducing the number of calls by 90% and consequently the chance of hitting the rate limit during list sync.

Both my 1.5k+ anime and 2.5k+ manga lists went from guaranteed rate-limit after about 20 pages in half a minute-ish to successful population within a few seconds for the 4-6 pages on the sync screen.

### DIFF
--- a/src/_provider/AniList/list.ts
+++ b/src/_provider/AniList/list.ts
@@ -110,72 +110,65 @@ export class UserList extends ListAbstract {
       this.username = await this.getUsername();
     }
 
-    let query = `
-    query ($page: Int, $userName: String, $type: MediaType, $status: MediaListStatus, $sort: [MediaListSort] ) {
-      MediaListCollection(status: $status, type: $type, userName: $userName, sort: $sort, chunk: $page, perChunk: 500, forceSingleCompletedList: true ) {
-        hasNextChunk
-        lists { entries {
+    const fields = this.compact
+      ? 'progress media { id idMal }'
+      : `
+        status
+        startedAt {
+          year
+          month
+          day
+        }
+        completedAt {
+          year
+          month
+          day
+        }
+        repeat
+        score(format: POINT_100)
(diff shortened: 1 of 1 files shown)


lolamtisch:
The problem here is that 500 entries loading at the same time will be quite heave on the list view.

But it is possible to distinguishe between background and frontend requests.

https://github.com/MALSync/MALSync/blob/master/src%2F_provider%2FMangaBaka%2Flist.ts#L95

BrutuZ:
Good point, and thanks for the reference, made it a easy tweak :+1:

Also took the chance to dedupe the query, decided it was worth it over keeping the diff lines count minimal

Strommy0815:
That sounds good! I hope this fixes the rate-limit problem I have with my Lists.

lolamtisch:
I found another problem. Sorting is applied by list state, Meaning when you select "all" in the bookmarks of the popup, is the sorting not correct. 

Not sure what the best solution is. Probably if background use MediaListCollection (maybe throw an error if sorting is something other then default, in case it is used for something in the future) and if not use the normal query

BrutuZ:
What about throwing a `.sort()` on the flatMap?

https://github.com/MALSync/MALSync/blob/60dadb9aeeb420e0a5401fe4c7e9ce35d7866cb7/src/_provider/AniList/list.ts#L183

lolamtisch:
First I do not want to implement the different sortings client side. Additionally I think that wont work well with pagination either.

BrutuZ:
That's fair. How would I check whether to use the new query or the old `Pages` one, `this.modes.frontend`?

lolamtisch:
Yeah. I think so

PR #3689, opened 2026-03-11
Add visibility checking functions: checkVisibility and isInViewport for chibi domFunctions

Source-Dom:


gickass:
do you think this checkVisibility is better?
```
/**
   * Checks if the selected element is currently visible
   * @input Element - DOM element
   * @options Object - Optional visibility settings
   * @returns Boolean indicating if the element is visible
   * @example
   * $c.querySelector('#player').checkVisibility() // uses defaults
   * $c.querySelector('#player').checkVisibility({ opacityProperty: false }) // overrides
   */
  checkVisibility: (ctx: ChibiCtx, input: Element, options: CheckVisibilityOptions = { visibilityProperty: true, opacityProperty: true }) => {
    return input.checkVisibility(options);
  },
```
  
if you write .checkVisibility() in chibi with no parameter, it will automically use { visibilityProperty: true, opacityProperty: true },

and you can still write like .checkVisibility({opacityProperty: false}) in chibi, to override the default incase you need different setting on that checkVisibility

Source-Dom:
Yes idk if people will use it tho, that's what I wasn't sure about, cuz it's cool to give more options but will people ever use it. But I guess we can give the choice it doesn't really matter.

Source-Dom:
Send a pr to my fork to the branch if you want.

Source-Dom:
Up

PR #3714, opened 2026-03-18
Add support of Animesite.fr

animesitefr:


src/pages-chibi/implementations/AnimeSite/main.ts
```diff
@@ -0,0 +1,108 @@
...
+  name: 'AnimeSite',
+  domain: 'https://animesite.fr',
+  languages: ['French'],
+  type: 'anime',
+  urls: {
+    // Sync pages: https://animesite.fr/play/{cryptedId}-{slug}/{season}/{episode}
+    // Overview pages: https://animesite.fr/{cryptedId}-{slug}
+    match: ['*://animesite.fr/*'],
+  },
+  search: 'https://animesite.fr/search/{searchtermPlus}',
```
lolamtisch, line 14: {searchterm} is correct in your case

src/pages-chibi/implementations/AnimeSite/main.ts
```diff
@@ -0,0 +1,108 @@
+import type { ChibiGenerator } from '../../../chibiScript/ChibiGenerator';
+import { PageInterface } from '../../pageInterface';
+
+export const AnimeSite: PageInterface = {
```
lolamtisch, line 4: You forgot to change the name here. And additoinally you need to import the page in pages.ts

lolamtisch (changes requested):
All the anime db pages do not follow the Series with Season approach, as your page does. There is not really a way to map that. So adding overivew support, makes no sense here.
For the video page you will have to make sure every season is handled as a seperate entry (identifier) and that the title at least gets appended something like Season 2 (It is ok to just use the season number). And do not extract the title from the url.

And in general nothing is working.

PR #3753, opened 2026-03-26
fix: sadscans #3731

### LINKED ISSUE #3731: Sadscans domain change (.com → .net) breaks MALSync detection
### Describe the Bug

Sadscans recently changed its domain from .com to .net. Because of this, MALSync no longer recognizes the site and fails to detect manga pages or sync progress. The domain needs to be updated in the extension.

1. Open Sadscans (.net)
2. Try to use MALSync on any manga page
3. MALSync does not detect the site or sync progress

Image

Image

### Browser

Chrome, Edge

### Type

Extension

### URL

https://sadscans.net/

Fen1kks:
Sadscans domain change (.com to .net)

The `npm run test:headless` command will again return a “timeout” (Exit code: 1). This is not due to any issues with the test setup, but because the Playwright test browser attempting to access the site is immediately flagged as a bot by Cloudflare and blocked.

Fixes #3731

### DIFF
--- a/src/pages-chibi/implementations/Sadscans/tests.json
+++ b/src/pages-chibi/implementations/Sadscans/tests.json
@@ -1,6 +1,7 @@
 {
   "title": "Sadscans",
   "url": "https://sadscans.net/",
+  "offline": true,
   "testCases": [
     {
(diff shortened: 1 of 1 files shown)


Fen1kks:
The site owners will be adding a direct synchronization feature to the site. For this reason, I was asked not to add it.

PR #3800, opened 2026-04-12
Th3Anime Integration

Boxofbiscuits97:
https://th3anime.me/

`tests.json` should be setup solidly but the `main.ts` might need some work as TypeScript is not my forte. I believe most of the functions should work besides possibly `getTitle()` but I've had no way to actually build and test the extension since I'm having some issues when running `npm install` in my project, it just loops infinitely. Assistance with `style.less` is definitely going to be need to get the MAL buttons showing up on the site.

### DIFF
--- a/src/pages-chibi/implementations/Th3Anime/main.ts
+++ b/src/pages-chibi/implementations/Th3Anime/main.ts
@@ -32,6 +32,9 @@ export const Th3Anime: PageInterface = {
         .number()
         .run();
     },
+    uiInjection($c) {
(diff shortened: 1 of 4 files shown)


Boxofbiscuits97:
>but I've had no way to actually build and test the extension since I'm having some issues when running npm install in my project, it just loops infinitely.

Disregard this part, I've gotten the project working and able to build properly now

Source-Dom:
You should change to Draft for now. It sends a notification to the owner, every time you make a change, btw.

PR #3825, opened 2026-04-26
Update MangaLivreTV implementation for toonlivre.net

Jhoorodre:
This PR updates the existing MangaLivreTV implementation in-place to support toonlivre.net.

Scope is intentionally limited to the implementation and its tests.
No README changes, no generated list changes, and no unrelated files.


Source-Dom:
Use your own account for commits not some supposed "malsync bot"

Jhoorodre:
> Use your own account for commits not some supposed "malsync bot"

Done, I rewrote the PR commit to use my own account (Jhoorodre) and force-pushed the branch.

Source-Dom:
do `npm run lint:fix`

Jhoorodre:
> do `npm run lint:fix`

I ran npm run lint:fix and pushed the changes to the same PR branch. Only formatting changes were applied to MangaLivreTV main.ts.

lolamtisch:
Did I miss something? mangalivre still exists and for toonlivre I only get a error.

Jhoorodre:
> Did I miss something? mangalivre still exists and for toonlivre I only get a error.

They are in the process of migrating a WordPress site to this new one. They reactivated the old site because not all of the database content had been migrated. However, the old site does not update with new content—it only works with the content available up to that point. It is intended for users who want access to content that has not yet been migrated.

It’s possible that there is a geolocation block, accessible only via VPN, country Brazil.

image

lolamtisch:
The domain is not registred at all.

https://www.nslookup.io/domains/www.toonlivre.net/webservers/

Jhoorodre:
> The domain is not registred at all.
> 
> https://www.nslookup.io/domains/www.toonlivre.net/webservers/

Ok, I wasn’t expecting that one — what do we do in this case? Should we leave it on standby? I opened this PR a day after the site went live, so let’s see if the site owners sort this out over the next few days.

lolamtisch:
Yeah. Just let me know when it is is back

Jhoorodre:
> Yeah. Just let me know when it is is back

Hi @lolamtisch,

It seems there was a misunderstanding regarding the domain status. The site is actually live and resolving, but only on the naked domain (toonlivre.net).

The owners haven't configured the DNS record for the www subdomain yet, which is why the previous lookup failed. As you can see in the attached screenshot, the main domain is active and pointing to Cloudflare IPs.

Could you please re-evaluate the PR or remove the blocked label?

Thanks!

image

PR #3835, opened 2026-05-03
Add miruro.bz to Miruro domains

BharatYadav007:
Register the new official Miruro domain (https://www.miruro.bz) in the Miruro page config. This updates the domain list and the URL match patterns in src/pages-chibi/implementations/Miruro/main.ts and includes a comment noting the source of the new domain.


lolamtisch:
Already added by #3842

Issue #13, opened 2018-10-23
Suggestion: Adding an actual timer for Estimated Upcoming Episode in MAL 


AccelerateIzaya:
Instead of words, is it possible if there is an actual timer there?
![](https://cdn.discordapp.com/attachments/420451445209563147/500624397560840202/unknown.png)

I searched around and found a site where we can add timers.
https://www.timeanddate.com/clocks/freecountdown.html

Would be really awesome if something like that would be added. 💯

PR #3840, opened 2026-05-03
feat(chibi): add DMM TV support

MeGaNeKoS:
## Summary

Adds chibi support for DMM TV (`tv.dmm.com`) anime tracking.

This integration supports:

- DMM TV playback pages:
  - `/vod/playback/on-demand/?season=...&content=...`
- DMM TV detail/overview pages:
  - `/vod/detail/?season=...`
  - `/vod/detail/?season=...&content=...`
- Title, identifier, episode, overview URL, and image extraction
- Episode list parsing from visible overview page links
- DMM-specific MALSync UI placement and styling

The implementation uses page URL parameters, document title, visible DOM, and JSON-LD metadata. It does not rely on network request interception.

## Edge Cases Covered

- DMM TV has two navigation modes: direct full-page detail loads and SPA/modal-style detail transitions from cards. The lifecycle now retriggers when the detail header/category content renders, so MALSync can initialize after DMM finishes replacing the page content.
- Anime/non-anime detection no longer depends on the numeric category id. It checks the rendered anime category text and the detail title anime/year pattern.
- Season selection is handled from DMM's visible detail header:
  - marker-only labels are combined with `h1` (`転生したらスライムだった件` + `第4期`)
  - full season titles are used directly (`... 3rd season`)
  - full-width season digits are normalized (`Season２` -> `Season 2`)
- This avoids incorrectly matching later seasons as season 1/base entries when DMM separates the base title from the selected season label.

## Testing

Ran:

```sh
npx.cmd eslint src/pages-chibi/implementations/DMMTV/main.ts --quiet
npx.cmd stylelint src/pages-chibi/implementations/DMMTV/style.less -q
npm.cmd run build:webextension:firefox
$env:FILES='[src/pages-chibi/implementations/DMMTV/tests.json,src/pages-chibi/implementations/DMMTV/main.ts]'; npm.cmd run test:headless
```

The scoped DMM headless tests passed.

Note: DMM TV can depend on Japan access/login/subscription state, so the test file is marked `unreliable`.

### DIFF
--- a/src/pages-chibi/implementations/DMMTV/main.ts
+++ b/src/pages-chibi/implementations/DMMTV/main.ts
@@ -1,0 +1,110 @@
+import type { ChibiGenerator } from '../../../chibiScript/ChibiGenerator';
+import { PageInterface } from '../../pageInterface';
+
+const domain = 'https://tv.dmm.com';
+
+export const DMMTV: PageInterface = {
+  name: 'DMM TV',
+  domain,
+  languages: ['Japanese'],
+  type: 'anime',
+  urls: {
+    match: ['*://tv.dmm.com/vod/*'],
+  },
+  search: 'https://tv.dmm.com/vod/search/?keyword={searchterm}',
+  sync: {
+    isSyncPage($c) {
+      return $c
+        .and(
+          $c.url().urlPart(3).equals('vod').run(),
+          $c.url().urlPart(4).equals('playback').run(),
+          $c.url().urlParam('season').boolean().run(),
+          $c.url().urlParam('content').boolean().run(),
+        )
+        .run();
+    },
+    getTitle($c) {
+      return pageTitle($c)
+        .replaceRegex('\\s+第\\d+話.*$', '')
+        .replaceRegex('\\s+\\(アニメ/\\d{4}年\\).*$', '')
+        .trim()
+        .run();
+    },
+    getIdentifier($c) {
+      return $c.url().urlParam('season').string().run();
+    },
+    getOverviewUrl($c) {
+      return $c
+        .string(`${domain}/vod/detail/?season=`)
+        .concat($c.this('sync.getIdentifier').run())
+        .run();
+    },
+    getEpisode($c) {
+      return pageTitle($c).regex('第(\\d+)話', 1).number().run();
+    },
+    uiInjection($c) {
+      return $c.querySelector('body').uiAppend().run();
+    },
+  },
+  overview: {
+    isOverviewPage($c) {
+      return $c
+        .and(
+          $c.url().urlPart(3).equals('vod').run(),
+          $c.url().urlPart(4).equals('detail').run(),
+          $c.url().urlParam('season').boolean().run(),
+        )
+        .run();
+    },
+    getTitle($c) {
+      return getJsonData($c).get('name').string().trim().run();
+    },
+    getIdentifier($c) {
+      return $c.url().urlParam('season').string().run();
+    },
+    uiInjection($c) {
+      return $c.querySelector('#detail-header').ifNotReturn().uiAfter().run();
+    },
+    getImage($c) {
+      return getJsonData($c).get('image').string().ifNotReturn().run();
+    },
+  },
+  list: {
+    elementsSelector($c) {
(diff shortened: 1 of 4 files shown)


MeGaNeKoS:
Follow-up after manual testing on Firefox:

The DMM TV edge cases found during testing are fixed now, including SPA/modal card navigation and season-specific title handling (`第4期`, full season titles, and `Season２` -> `Season 2`).

I also rebuilt and retested the Firefox extension with these fixes. This should be ready for review/merge.

src/pages-chibi/implementations/DMMTV/main.ts
```diff
@@ -0,0 +1,200 @@
...
+        )
+        .domReady()
+        .trigger()
+        .run();
+    },
+    overviewIsReady($c) {
+      return $c.waitUntilTrue($c.querySelector('h1').boolean().run()).trigger().run();
+    },
+    listChange($c) {
+      return $c.detectChanges($c.querySelector('main').text().run(), $c.trigger().run()).run();
```
lolamtisch: Triggers on all websites all the time.

lolamtisch (changes requested):

src/pages-chibi/implementations/DMMTV/main.ts
MeGaNeKoS: Thanks, addressed.

The broad `listChange` watcher has been removed for DMM TV, so it no longer watches `main.text()` or triggers list handling repeatedly. SPA support is now handled by narrower DMM-specific readiness triggers (`detectURLChanges`, title changes, detail-header h1/season-label count, and DMM category-link count).

I also removed text-based `detectChanges` reads, added early returns for missing title/episode values, and covered the rated detail route plus English `Episode N` labels.

Tested with scoped DMM headless tests and Firefox extension build.

MeGaNeKoS:
One DMM edge case surfaced a separate core limitation around local corrections: DMM can expose one shared `season` identifier while continuing episode numbers across multiple MAL seasons, e.g. `#1-12` for Season 1 and `#13+` for Season 2.

The current DMM implementation keeps returning the raw DMM episode number, which lets the existing rules engine handle episode-range remaps when a central rule exists. I opened a separate issue for the missing local/user-editable range-rule support instead of trying to solve it inside this DMM page implementation:

https://github.com/MALSync/MALSync/issues/3894

MeGaNeKoS:
﻿Small follow-up from testing U-NEXT as well: DMM TV has the same direct-player limitation.

For DMM, URLs like:

`/vod/playback/on-demand/?season=...&content=...`

are stable enough for direct player detection, but the URL shape itself does not prove whether the title is anime. Non-anime DMM titles can use the same playback URL structure, so direct player support can false-positive if we only look at the playback URL.

I handled U-NEXT similarly in #3918 so direct player pages work, and opened #3917 to discuss a reusable core/site metadata approach for this class of direct-player detection.

lolamtisch:
I just looked into the anime concern. As far as I can see, does the `operationName: "FetchVideo"` call contain the categories. So you should be able to get the info with the requestProxy (It listens for all requests the page makes). Maybe with that you do not need to use that many fallbacks everywhere too.


https://github.com/MALSync/MALSync/blob/master/src/pages-chibi/implementations/Plex/main.ts#L16

image

MeGaNeKoS:
﻿Thank you for the implementation idea. I checked the DMM TV player/detail network flow again and confirmed your suggestion.

The page-made `FetchVideo` GraphQL response includes `video.categories`. For anime it returns:

```json
"categories": [
  { "id": "15", "name": "???" }
]
```

and for non-anime it returns other categories, for example `DMM TV?????` / `?????`.

I updated the DMM implementation to follow your suggestion and use Chibi `requestProxy` for that `FetchVideo` response. The implementation now caches the video metadata by `DMM TV#<seasonId>` only when the response contains the anime category, and overview/player detection requires that cached anime metadata before triggering.

I think the DMM implementation should be complete now. Validation so far:

- `npm run build:webextension:content` passes.
- ESLint on the DMM implementation has no errors, only existing warnings.

lolamtisch (approved):

PR #3864, opened 2026-05-11
Comix next chapter & chapter highlighting

Yumeo0:
Adds support for next chapter & chapter highlighting to comix.to


kavillock:
ok, i wave a question, why you rewrite JSON data extraction to DOM/URL based, this data are there, but selector changed (from #syncData to #initial-data)

Yumeo0:
> ok, i wave a question, why you rewrite JSON data extraction to DOM/URL based, this data are there, but selector changed (from #syncData to #initial-data)

Idk. Because the current code didn't work and I tried to fix it somehow. It's the first time I work with this codebase and the wiki didn't explain much.

PR #3882, opened 2026-05-15
tsukuyomi , remove getmalurl, fix identifier

gickass:
tsukuyomi just change their site a bit it broke getmalurl


henrik9999:
The identifier should be just the random id part after the last -

gickass:
i can take anilist id from [property="og:image"] but maybe it is bad idea

lolamtisch (approved):

PR #3891, opened 2026-05-17
fix(SakuraMangas): update UI injection and enable quicklinks

Jhoorodre:
- Updated overview uiInjection to use .groupbutton selector
- Added search property to enable AniList quicklinks support

### DIFF
--- a/src/pages-chibi/implementations/SakuraMangas/main.ts
+++ b/src/pages-chibi/implementations/SakuraMangas/main.ts
@@ -8,6 +8,7 @@ export const SakuraMangas: PageInterface = {
   urls: {
     match: ['*://*.sakuramangas.org/*'],
   },
+  search: 'https://sakuramangas.org/',
   sync: {
     isSyncPage($c) {
       // Reading page follows pattern: /obras/{slug}/{number}/
@@ -59,8 +60,8 @@ export const SakuraMangas: PageInterface = {
       return $c.this('sync.getIdentifier').run();
     },
     uiInjection($c) {
-      // Inject UI after the first col-xxl-10 element
-      return $c.querySelector('.col-xxl-10').uiAfter().run();
+      // Inject UI after the first groupbutton element
+      return $c.querySelector('.groupbutton').uiAfter().run();
     },
     getImage($c) {
       return $c
--- a/src/pages-chibi/implementations/SakuraMangas/main.ts
+++ b/src/pages-chibi/implementations/SakuraMangas/main.ts
@@ -8,7 +8,7 @@ export const SakuraMangas: PageInterface = {
   urls: {
     match: ['*://*.sakuramangas.org/*'],
   },
-  search: 'https://sakuramangas.org/',
+
   sync: {
     isSyncPage($c) {
       // Reading page follows pattern: /obras/{slug}/{number}/


src/pages-chibi/implementations/SakuraMangas/main.ts
```diff
@@ -8,6 +8,7 @@ export const SakuraMangas: PageInterface = {
   urls: {
     match: ['*://*.sakuramangas.org/*'],
   },
+  search: 'https://sakuramangas.org/',
```
lolamtisch: https://sakuramanga.org/?s={searchtermPlus}

src/pages-chibi/implementations/SakuraMangas/main.ts
```diff
@@ -59,8 +60,8 @@ export const SakuraMangas: PageInterface = {
       return $c.this('sync.getIdentifier').run();
     },
     uiInjection($c) {
-      // Inject UI after the first col-xxl-10 element
-      return $c.querySelector('.col-xxl-10').uiAfter().run();
+      // Inject UI after the first groupbutton element
+      return $c.querySelector('.groupbutton').uiAfter().run();
     },
```
lolamtisch, line 65: Please Remove uiInjection from sync and add overview page support.

https://sakuramanga.org/p-toriko/

lolamtisch (changes requested):

Jhoorodre:
@lolamtisch 

I think there was a mix-up. The fix I pushed is for Sakuramangas, with an “S”, while the example you gave is Sakuramanga, without the “S”.

By the way, regarding the quick links on sakuramanga.org, they use a floating modal and direct API requests, which I don’t think MAL Sync supports. I mainly added the quick links so they’re visible for anyone using it—it’s pretty common for me to forget this extension even exists lol.

Domain: https://sakuramangas.org/

src/pages-chibi/implementations/SakuraMangas/main.ts
Jhoorodre: Do you think we can do something with this?

https://sakuramangas.org/dist/sakura/global/sidebar/sidebar.core.php?q=tower%20dungeon

src/pages-chibi/implementations/SakuraMangas/main.ts
Jhoorodre: The real work overview page on the site:

https://sakuramangas.org/obras/tower-dungeon/

lolamtisch:
Ah yeah you are right. The points still hold tho. If there is an overview page the ui injection has to go on that. 

And I can not find a search url either. So it has to be removed.

Jhoorodre:
> Ah yeah you are right. The points still hold tho. If there is an overview page the ui injection has to go on that. 
> 
> And I can not find a search url either. So it has to be removed.

The UI injection was already done on the overview page; I actually made this fix to resolve the issue that was broken. Could you please double-check that?

Regarding the search URL: is there any way to implement a quicklink without it? Without that URL, the site doesn't appear in external sources, so I'm trying to find a way to maintain that functionality.

lolamtisch:
If the page does not have a url where the serach can be encoded in. Then there is nothing we can do. The external sources not showing, should already be fixed.

PR #3923, opened 2026-05-26
Animetsu Commit

Cru-ton:
Adding animetsu to the list of the pages


henrik9999:
Animetsu is already implemented

PR #3972, opened 2026-06-09
Fix slow loading of Shikimori list

### LINKED ISSUE #3601: List sync with Shikimori is taking too long
A list of 607 entries is processed in one and a half minutes, sending a bunch of requests for 25 entries and reaching the rate limit. In the v1 API, there is a method https://shiki.one/api/doc/1.0/users/anime_rates that returns all the data from the list in one request, including titles, which are not in the v2 API, if these requests are due to their absence.

Image

selpats:
Closes #3601


selpats:
With my list, the loading speed became a few seconds instead of a minute.

selpats:
@lolamtisch Could you please review this? Sorry for not providing any info right away, but because the rate limits were not being handled, list synchronization was taking way too long. The rate limits are specified here: https://shikimori.io/api/doc/2.0 (5 requests per second and 90 per minute). Also, the correct limit for fetching title information is documented here: https://shikimori.io/api/doc/1.0/animes/index (50 instead of 25). I'm not sure if this is the best approach for handling rate limits, since I'm not a programmer and I figured it out with Gemini, but the code works.

PR #4018, opened 2026-06-26
[TASK] Add Mkissa page

Bissous:
Adds a page for mkissa.to (manga). It runs on the same allanime backend as allManga but the frontend is a different Svelte app, so I couldn't reuse allManga's selectors and had to write new ones.

The overview page (/manga/<id>) grabs the title and cover and injects the UI. I also pull the AniList id out of the banner image so AniList users get matched directly, otherwise it falls back to searching by title. The reader (/manga/<id>/chapter-N-sub) takes the series name from the page title, the chapter number from the URL, and the page count from the reader. Deep links are behind Cloudflare and the reader is an SPA route, so it only syncs once you're moving around inside the site. I watch document.title instead of the URL so it re-syncs when you switch chapters. Left anime out for now, that player has its own Cloudflare check, can do that later.

Tested: build passes, lint and the ts tests pass. I checked the real output against actual pages from the site (an overview and a chapter), page detection, title, id, episode, overview url, image, UI injection, the AniList url and the reader progress all come out right. What I didn't do is run the extension live on the site clicking through chapters, I couldn't get a clean run past Cloudflare for that. I did confirm document.title changes on every chapter though, which is what the title-watching relies on.

### DIFF
--- a/src/pages-chibi/implementations/Mkissa/style.less
+++ b/src/pages-chibi/implementations/Mkissa/style.less
@@ -1,7 +1,10 @@
 @import './../pages';
 
-@boxBackground: #ffffff14;
-@boxHighlightBackground: #f59e0b;
-@boxHighlightFontColor: #000;
+@boxBackground: var(--color-surface-muted) !important;
+@boxFontColor: var(--color-text-secondary);
+@boxHighlightBackground: var(--color-manga-accent) !important;
+@boxHighlightFontColor: var(--color-manga-surface) !important;
 
-@activeEp: #f59e0b40;
+.box() {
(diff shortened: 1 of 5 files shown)


src/pages-chibi/implementations/Mkissa/style.less
```diff
@@ -0,0 +1,8 @@
+@import './../pages';
```
lolamtisch, line 1: None of the variables seem to exist for me

lolamtisch (changes requested):

Bissous:
those vars came from the Comix theme, replaced them with fixed values.

lolamtisch (approved):

PR #4043, opened 2026-07-03
[BUGFIX] Fix Comix progress tracking and overview UI #4034

### LINKED ISSUE #4034: Comix bugs
### Describe the Bug

Hello!

I noticed two bugs on Comix:

1. When logged into Comix, the website integration for a given manga (showing score, status, chapter number, etc.) does not appear unless you go back in the browser and then forward into the manga page again. But when you log out of Comix, the integration appears properly.
2. In the reader, when the progress bar is disabled, MALSync immediately marks the chapter as read even when "Sync manga after reading 90% of the pages" is enabled. But when the progress bar is enabled, the chapter is marked as read only when the end of the chapter is reached, as expected.

If you could look at these issues, that would be great! Thank you.

1. Log into Comix and choose a manga. The MALSync integration should appear, but does not.
2. Open any manga chapter on Comix, and disable Comix's progress bar from the options menu.

### Browser

Firefox

### Type

Extension

### URL

https://comix.to

Bissous:
Two separate issues on Comix:
Chapters get marked as read instantly when the reader's progress bar is turned off. Comix removes `.rpage-progress` from the DOM in that mode, so the readerConfig ended up at 0/0 and tracking fell back to instant. Added a condition on that config plus a fallback that counts the `.rpage-page` elements instead.
The overview UI doesn't appear when logged in unless you navigate back and forward. Title and id come from the `#syncData` JSON, which is ready before the page finishes rendering, so the UI was injected too early into a missing target. It now uses the site's own `#mal-sync` slot and waits for it.

Tested on comix.to with the progress bar both on and off, progress reads correctly at every scroll position. Overview injection into `#mal-sync` also checked.
Not tested: the logged-in overview case on a real account, I only reproduced the timing with a simulated injection, so a second check there would be welcome.

Fixes #4034

### DIFF
--- a/src/pages-chibi/implementations/Comix/main.ts
+++ b/src/pages-chibi/implementations/Comix/main.ts
@@ -99,6 +99,12 @@ export const Comix: PageInterface = {
         .ifNotReturn($c.trigger().return().run())
         .run();
     },
+    overviewIsReady($c) {
+      return $c
+        .waitUntilTrue($c.querySelector('.mpage__desc-wrap').boolean().run())
+        .trigger()
+        .run();
+    },
     listChange($c) {
       return $c
         .detectChanges(


src/pages-chibi/implementations/Comix/main.ts
```diff
@@ -59,7 +66,7 @@ export const Comix: PageInterface = {
       return $c.querySelector('.mpage__poster img').getAttribute('src').ifNotReturn().run();
     },
     uiInjection($c) {
-      return $c.querySelector('.mpage__desc-wrap').uiBefore().run();
+      return $c.querySelector('#mal-sync').uiAppend().run();
```
lolamtisch: I feel like the previous place is better or not?

src/pages-chibi/implementations/Comix/main.ts
```diff
@@ -36,13 +36,20 @@ export const Comix: PageInterface = {
...
         current: $c =>
           $c
             .querySelectorAll('.rpage-progress__seg.is-visited, .rpage-progress__seg.is-active')
             .length()
             .run(),
         total: $c => $c.querySelectorAll('.rpage-progress__seg').length().run(),
       },
+      // The progress bar is removed from the dom when disabled in the reader settings,
+      // so fall back to counting the page wrappers.
+      {
```
lolamtisch: Does not work and on top breaks fallback to instant update

lolamtisch (changes requested):

Bissous:
Thanks for taking a look, both fair points.

Yeah, the reader progress fallback was broken. Counting the page wrappers only works in the long-strip reader, where it matches the actual page. In the paged reader it returns 0, and because the total was still non-zero it never fell back to instant update like it should have. I've added a condition so it only runs for the long-strip reader now, and everything else falls through to instant update the way it did before. I compared the wrapper count against the site's own segment counter on a long-strip chapter to be sure, and it follows the current page correctly.

For the overview, you're right, I moved it back to .mpage__desc-wrap. The real problem when logged in was just timing: the title and id come from the #syncData JSON, which loads before the page is actually rendered, so the default check passed and the UI got injected into nothing. overviewIsReady now waits for .mpage__desc-wrap to show up before injecting, so it works logged in without moving the UI or depending on #mal-sync.

Both changes are in the two commits above.

src/pages-chibi/implementations/Comix/main.ts
```diff
@@ -36,13 +36,22 @@ export const Comix: PageInterface = {
...
             .querySelectorAll('.rpage-progress__seg.is-visited, .rpage-progress__seg.is-active')
             .length()
             .run(),
         total: $c => $c.querySelectorAll('.rpage-progress__seg').length().run(),
       },
+      // The progress bar is removed from the dom when disabled in the reader
+      // settings. The long-strip reader still lets us track progress from the
+      // page wrappers; other reader modes fall through to instant tracking.
+      {
+        condition: $c => $c.querySelector('.rpage-main--long-strip').boolean().run(),
```
lolamtisch: I dont think this is a good solution. Because now we have inconsistencies. Specially because I imagine it switches to long stripe and paged view depending on manga/manhwa (did not test if actually the case).

It is just easier to tell the people to just enable the progressbar if they want page tracking. It has a lot less ambiguities.

lolamtisch (changes requested):

src/pages-chibi/implementations/Comix/main.ts
Bissous: Yeah, that's fair. Since the reader switches between long strip and paged depending on the title, page tracking would only work part of the time, which is worse than just being consistent. I've dropped the reader change entirely, so that part stays as it was and enabling the progress bar remains the way to get page tracking.

The PR is now just the overview timing fix.

lolamtisch (approved):

PR #4045, opened 2026-07-04
Add setting for finished release progress

llccrr:
Finished entries currently hide release progress because complete release data makes `ProgressRelease.isAiring()` false. That can be annoying when a show finishes airing while the user is behind: miniMAL falls back to only showing the watched count, so it is no longer obvious that released episodes are still unwatched.

This adds an opt-in setting for finished entries. It does not change release-progress fetching. If existing release progress data marks an entry as finished and `totalEp > watchedEp`, miniMAL uses the entry total as the release progress. For example, `6/12` shows the red availability bar up to `12`, while the blue watched bar still covers the first `6`.

Default behaviour stays unchanged.

Tested:
- `git diff --check`
- `eslint src/utils/progressRelease.ts src/_minimal/views/bookmarks.vue src/_minimal/components/overview/overview-image.vue src/_minimal/components/overview/overview-update-ui.vue` (warnings only)
- locale JSON parse
- `node webpackConfig/webextension.content.generateChibi.js`

Not tested:
- `vue-tsc --noEmit` still fails on existing `ChibiGeneratorTypes.ts` errors unrelated to this change.

### DIFF
--- a/assets/_locales/en/messages.json
+++ b/assets/_locales/en/messages.json
@@ -1267,6 +1267,10 @@
         "message": "Load PTW/PTR for Progress evaluation",
         "description": "In addition to CW/CR also load PTW/PTR when checking Progress"
     },
+    "settings_progressShowFinished": {
+        "message": "Show release progress for finished entries",
+        "description": "Option to show release progress bars and pills for finished anime or manga entries"
+    },
     "settings_Video_Player": {
         "message": "Video player",
         "description": "Video settings header"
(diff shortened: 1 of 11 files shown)


henrik9999:
We don't do that intentionally that will send billion more requests to our api

Edit: ok maybe I am wrong about API requests 

Edit2: but if I remember correctly we only update airing entries over API so it wouldn't show the new episodes probably

llccrr:
Thanks for checking this! I simplified/tightened the PR so it does not change fetch behavior.

The new behavior only uses release progress that is already loaded/cached. When release data is marked complete and the entry has `totalEp > watchedEp`, miniMAL displays the release progress as the total episode count. So for `6/12`, the red availability bar fills to `12` and the blue watched bar still covers `6`.

If there is no existing release-progress data for that finished entry, this does not request anything extra or show new release progress.

PR #4069, opened 2026-07-12
fix(pages-chibi): detect Jellyfin anime tagged only on the series

wobkobi:
Jellyfin only puts the `Anime` genre/tag on the Series item. The episodes and seasons come back with empty genres and tags, so `isAnime` always returned false on an episode and nothing ever synced.

On top of that, `checkMetadataRequest` was ignoring the Series response entirely (it only handled Season/Episode/Movie), so the one response that actually had the anime marker was being thrown away.

Fix:
- Handle the `Series` type and remember whether a series is anime, keyed by its id.
- Episodes and seasons now count as anime if their own fields say so, or if their parent series was already flagged. If the series hasn't been seen yet, the episode gets stashed and flushed once the series response comes in.

So it works regardless of which response arrives first (e.g. landing directly on an episode). Movies are unchanged, and no new chibi features are used.

Tested against a real server (ZENSHU) where the anime tag is only on the series.

### DIFF
--- a/src/pages-chibi/implementations/Jellyfin/main.ts
+++ b/src/pages-chibi/implementations/Jellyfin/main.ts
@@ -30,7 +30,11 @@ type MovieMetadata = {
   Name: string;
 } & CommonMetadata;
 
-type Metadata = EpisodeMetadata | SeasonMetadata | MovieMetadata;
+type SeriesMetadata = {
+  Type: 'Series';
+} & CommonMetadata;
+
+type Metadata = EpisodeMetadata | SeasonMetadata | MovieMetadata | SeriesMetadata;
 
 type OverviewMeta = SeasonMetadata;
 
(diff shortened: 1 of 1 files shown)


lolamtisch (commented):
It has the problem I suspected. When starting an Episode from the home page, then the series information never gets parsed.

As it is they way most people watch episodes. It is kind of a dead end. As far as I see the parsing the folder path is the only reliable way to check for anime.

lolamtisch:
An option would be to implement a permament variable using the websites localstorage. But still a bit janky.

wobkobi:
using the folder path isn't good for me personally so I'm going use localstorage instead.

lolamtisch:
In that case the best way would to create a seperate PR for the extension of chibi. That way it can be merged independent. 

The actuall jellyfin PR can only be merged when the new chibi feature is part of the next release and needs to stay open until then.

wobkobi:
kk

wobkobi:
#4131

wobkobi:
https://github.com/MALSync/MALSync/pull/4131 has now been merged. Can this be looked at again?

PR #4072, opened 2026-07-12
[BUGFIX] Ignore key auto-repeat in the shortcut listener

### LINKED ISSUE #3978: "Correct Anime" popup closes/disappears instantly on streaming sites (focus loss bug)
### Describe the Bug

When attempting to use the "Correct Anime" feature (using the keyboard shortcut or clicking the UI link) to correct a misidentified anime season, the overlay pop-up appears for only a split second and instantly disappears/closes. 

This happens before the user has any time to type, paste a correct MyAnimeList link, or interact with the menu. All other extensions have been deactivated to rule out conflicts, confirming the bug is isolated to the MAL-Sync interface element when losing focus immediately.
https://github.com/user-attachments/assets/25e7d59c-b65a-4066-8989-d3515104eb8f

1. Go to an anime streaming site page where a title or season is misidentified (e.g., streaming Season 4 but tracked as Season 1).
2. Trigger the anime correction window (e.g., press Alt + O on the physical keyboard).
3. Observe the correction popup flash on the screen for roughly 1 second or less before automatically closing itself.

### Browser

Google Chrome

### Type

Extension

### URL

https://anime-sama.to/

Bissous:
Fixes #3978

Managed to reproduce this on anime-sama.to. Despite the issue title it's not a focus problem there is no blur handler anywhere that could close the popup.

What's actually going on: `openCorrection()` is a toggle (calling it while the popup is open closes it), and the shortcut listener fires on every keydown, including the auto-repeated ones you get from holding a key. Hold the shortcut slightly too long and the popup opens and closes in a loop, which looks exactly like the "flashes then disappears" from the video.

Single-key shortcuts (the default is just `C`) have a sneakier variant: `checkShortcut` requires the exact number of pressed keys, so while a modifier is held the shortcut doesn't fire at all. It ends up firing on the modifier's keyup instead, and the first auto-repeat of the key you're still holding closes the popup right after it opened.

One physical press should mean one trigger, so the fix is to ignore keydown events that have `e.repeat` set.

Chrome (unpacked build) on anime-sama.to. Before: holding the shortcut key toggled the popup open/closed repeatedly (5 keydowns = 5 toggles). After: it opens once and stays open. A deliberate second press still closes it, that behavior is unchanged. Not tested on Firefox.

### DIFF
--- a/src/utils/player.ts
+++ b/src/utils/player.ts
@@ -271,8 +271,6 @@ export function shortcutListener(callback) {
     function keyEvent(e) {
 
       e = e || event;
-      // Holding a key fires repeated keydowns, each re-triggering the shortcut
-      // (e.g. toggling the correction popup open/closed).
       if (e.repeat) return;
       const key = e.which || e.keyCode;
       keyMap[key] = e.type === 'keydown';
--- a/src/utils/player.ts
+++ b/src/utils/player.ts
@@ -271,6 +271,9 @@ export function shortcutListener(callback) {
     function keyEvent(e) {
 
       e = e || event;
+      // Holding a key fires repeated keydowns, each re-triggering the shortcut
+      // (e.g. toggling the correction popup open/closed).
+      if (e.repeat) return;
       const key = e.which || e.keyCode;
       keyMap[key] = e.type === 'keydown';
 


src/utils/player.ts
```diff
@@ -271,6 +271,9 @@ export function shortcutListener(callback) {
     function keyEvent(e) {
 
       e = e || event;
+      // Holding a key fires repeated keydowns, each re-triggering the shortcut
+      // (e.g. toggling the correction popup open/closed).
+      if (e.repeat) return;
```
lolamtisch:
```suggestion
      if (e.repeat) return;
```

lolamtisch (changes requested):

lolamtisch (approved):
Good catch. Thank you

PR #4073, opened 2026-07-12
[AnimeGO] fix new css selectors and name detection

NoPlagiarism:
- fix css selectors after ui update
 - fix title detection (why they removed it.. hate my life)
 - adjusted tests


lolamtisch:
Does the page still exist?

NoPlagiarism:
Yes, it does

image

PR #4095, opened 2026-07-28
Fix issues with JKanime functionality

36095:
MAL-Sync was broken on JKanime due to last year's changes in their UI, this PR adds compatibility with those UI changes.

- `src/pages/jkanime/main.ts`
  - updated the selectors that were broken
- `src/pages/jkanime/style.less`
  - updated styles to adapt to new UI
- `src/pages/jkanime/meta.json`
  - updated urls to exclude some new urls that were being detected incorrectly
  - removed a url exclusion that were no longer in use by the site

### DIFF
--- a/src/pages/Jkanime/main.ts
+++ b/src/pages/Jkanime/main.ts
@@ -21,10 +21,9 @@ export const Jkanime: pageInterface = {
   sync: {
     getTitle(url) {
       return j
-        .$('div.breadcrumb-option > div > div > div> div.breadcrumb__links > h1')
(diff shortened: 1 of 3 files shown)


lolamtisch:
We do not accept changes to legacy implementations. You will need to implement it as an Chibi website and remove the old implementation.

36095:
oh, got it, didn't see that, then i'll take a look at that and change/add it that way.

I suppose i'll close this PR since it doesn't fit, thanks. 😄

PR #4109, opened 2026-07-30
Fixed implementation for JKAnime

36095:
MAL-Sync was broken on JKanime due to last year's changes in their UI.

This PR adds a new Chibi page implementation to support syncing anime progress from JKAnime, and registers it in the Chibi pages registry, adding compatibility with the changes made to their UI so it can be detected and loaded correctly by MAL-Sync.

**Changes:**
- Removed legacy implementation
- Registered a new `JKAnime` page in `src/pages-chibi/pages.ts`.
- Added the JKAnime Chibi implementation (`main.ts`) with sync, lifecycle, and search URL.
- Added basic test fixtures (`tests.json`) and shared styling import (`style.less`) for the new implementation.

Tested on multiple overview/sync pages, with and without: autoplay, fullscreen and autoresume.
Tested using only MyAnimeList sync bc I don't use the other ones.

> P.S: I already did this fix a couple of days ago, but using the legacy implementation, thanks to [lolamtisch](https://github.com/lolamtisch) for letting me know I had to use the new system ❤️.


src/pages/list.json
henrik9999, line 1: Please don't modify this file it's autogenerated

36095 (commented):
Oh, didn't know that, sorry, fixing rn

src/pages-chibi/implementations/JKanime/main.ts
```diff
@@ -0,0 +1,128 @@
...
+    },
+    listChange($c) {
+      return $c
+        .detectChanges(
+          $c
+            .querySelectorAll('#episodes-content > div.epcontent')
+            .last()
+            .find('a')
+            .getAttribute('href')
+            .run(),
```
lolamtisch: Error when does not exist. Needs .ifNotReturn()


Image


```suggestion
            .querySelectorAll('#episodes-content > div.epcontent')
            .ifNotReturn()
            .last()
            .find('a')
            .ifNotReturn()
            .getAttribute('href')
            .run(),
```

lolamtisch (changes requested):
I know the old implementation did use them too. Could you go through all the selectors and improve them? Like this they will break with any mirror page change.

src/pages-chibi/implementations/JKanime/main.ts
36095: working on it rn, if there's something else that needs attention, just tell me :)

36095: Resolved in commit [6455caf](https://github.com/MALSync/MALSync/pull/4109/commits/6455cafc1477e17f24074dc0fc5948d375561966) (and [3691c9c](https://github.com/MALSync/MALSync/pull/4109/commits/3691c9c040a08f17e8c4d3d9a27901586e00b1b0))
For what I can tell it is resolved, again, if there's something else, just tell me.

Initially it was trying to get the list elements in the episode pages, although there IS a list in those pages they use different selectors, so i'll try to make it work for both first, so this is kind of a temporary fix (if I can get it to work 😅).

36095:
It took me a while to make it use the minimum amount of selectors possible, but now instead of flooding the script with selectors, we construct the most part using less breakable things like the url of the page, as well as the remaining selectors were simplified so they doesn't break that easily

now the implementation should do the following:

Sync:

- `getTitle()`
  - now gets anime name from the page title instead of using selectors
- `getImage()`
  - now gets the image constructing their cdn url directly instead of using selectors
- `getOverviewUrl()`
  - now uses `sync.getIdentifier` with `urlAbsolute()` to construct the overview page URL
- `nextEpisodeUrl()`
  - the selector was simplified to be more flexible

Overview:

- `isOverviewPage()` now uses the page title instead of an arbitrary selector to decide wether it is an overview page or not
- `getTitle()`
  - now gets anime name from the page title instead of using selectors
- `getImage()`
  - now gets the image constructing their cdn url directly instead of using selectors

List:

- `elementSelector()`
  - although both sync and overview pages have episode lists, now it was made to only work in overview pages because it was giving some errors on sync pages

Lifecycle:

- `listChange()`
  - now it only checks in overview pages
  - now it can handle `404` http errors and returns when it gets one

Tests:

- Added a new test case for when there's no next episode

Styles:

- Adjusted some styles' look & feel
- Made the font colors adapt to light and dark modes
- Removed sync pages episode list styles that doesn't fit anymore

----

If there's anything else that I missed, please let me know

lolamtisch:
Still the selectors could be less specific + quite specific expressions

36095:
> Still the selectors could be less specific + quite specific expressions

I'll try to make the selectors less specific, but can you elaborate in the 'expressions' part pls?

lolamtisch:
With specific expressions I pretty much meant the tightly coupled string manipulations, which actually are worse as they fail silent in comparision to the selectors.

Issue #2145, opened 2023-11-18
Backup Settings


mariodac:
### Suggestion

Back up the settings to be exported to a file so they can be imported to another device

SiTWulf:
Any new about this feature?

hamzaharoon1314:
Hi, is there any update on this feature? It's really necessary, as it's challenging to readjust everything repeatedly, especially when using different browsers.

wipe-tartly-polar:
I am also requesting this feature!

SiTWulf:
It's been almost three years with no news about this request... Any updates?

Ash1421:
> It's been almost three years with no news about this request... Any updates?

> [!IMPORTANT]
>  I am currently working on a fork that the PR, will have.

### feat: settings search, about page, and backup system
Closes #2145, references #116

## Changes

**Settings search** — \`manage_search\` icon in the root settings view works in both the popup and the PWA. Fuzzy + substring match, keyboard navigation, respects existing \`condition()\` and \`system\` guards on each setting.

**About page** — live GitHub API data: version vs latest, install links, repo stats, contributor grid, creator credit, and backup status per provider.

**Backup & Restore** — local file export/import, Google Drive (chrome.identity + drive.appdata scope), WebDAV/NAS (Nextcloud, Synology, QNAP), Backblaze B2. Credentials in \`settings/backup_*\` keys (sync storage); timestamps in \`backup/lastRun_*\` (local storage)."

> [!NOTE]
>  This is a sneak peak at the features via an unpacked versions screenshots.
<details><summary>Screenshots</summary>
<p>

<img width="1919" height="976" alt="Image" src="https://github.com/user-attachments/assets/dab91e7c-9f7a-4ca4-bb8d-0b5a46485a63" />

<img width="1919" height="976" alt="Image" src="https://github.com/user-attachments/assets/1417abe3-27d1-4d90-97fe-b0dbdce6a58e" />

<img width="1919" height="976" alt="Image" src="https://github.com/user-attachments/assets/9a5371ba-0116-4698-ac3e-18b50d6924ac" />

<img width="1919" height="976" alt="Image" src="https://github.com/user-attachments/assets/2f65921e-3c8f-4096-b74e-ab7119719562" />

<img width="1919" height="976" alt="Image" src="https://github.com/user-attachments/assets/2b4b2695-0a56-4901-8b4a-3b0b771d6f98" />

</p>
</details> 

> [!WARNING]
[...]

PR #4154, opened 2026-08-19
feat: add settings search bar

Ash1421:
Adds a manage_search icon at the top of the root settings view. Works in both the popup and the PWA. Filters the full settings tree in real time with fuzzy + substring match and match highlighting. Keyboard: arrows to navigate results, Enter to jump, Ctrl+F to open, Escape to close. Respects existing condition() and system guards.

Related: #2145

### DIFF
--- a/src/_minimal/components/settings/settings-search.vue
+++ b/src/_minimal/components/settings/settings-search.vue
@@ -1,0 +1,355 @@
+<template>
+  <div class="settings-search" :class="{ active: open }">
+    <div class="search-bar" @click="handleBarClick">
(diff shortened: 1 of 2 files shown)


lolamtisch:
Seems a bit detached from the remaining codebase

Ash1421:
> Seems a bit detached from the remaining codebase

Could you elaborate please ?

lolamtisch:
It does not integrate well into the existing codebase

Ash1421:
> It does not integrate well into the existing codebase

And how does it not "integrate" "well"

Please tell me how this wouldn't be a good feature to add

Issue #116, opened 2019-06-15
Save local data into google drive for a more safe backup.


manolito62:
IN case the comp'uter or the HDD gets broken, the date would be safe on a cloud, the local data would be sdafe on the cloud at certain interval.

andku85:
well should have option to backup all to google drive as well as having an active local saving to google drive. vs using the small cache amount in the extension.

crse:
Related to #2145

PR #4155, opened 2026-08-19
feat: add Backup & Restore settings

### LINKED ISSUE #2145: Backup Settings

Ash1421:
Closes #2145, references #116

Four providers:
- Local file: full chrome.storage snapshot as a dated JSON file
- Google Drive: launchWebAuthFlow OAuth (cross-browser: Chrome, Firefox, Opera, Edge), drive.appdata scope, user-provided client ID
- WebDAV / NAS: Nextcloud, Synology, QNAP, any WebDAV server
- Backblaze B2: native B2 API, SHA-1 integrity, no external SDK

Restore disabled until connection tested this session. Export shows a security notice that the file contains credentials. Credentials in settings/backup_* keys (sync storage, syncs across browsers). Timestamps in backup/lastRun_* (local storage, per-device).

Adds form-password.vue: password input with show/hide eye toggle. Hidden in userscript mode (chrome.storage not available there).

Google Drive requires an OAuth 2.0 client ID from Google Cloud Console with drive.appdata scope. See PR description for setup.

> [!NOTE]
> Google Drive setup (for maintainer or self-hosting)
>
>Go to [Google Cloud Console → Credentials](https://console.cloud.google.com/apis/credentials)
>Create a project → Enable Google Drive API
>Create an OAuth 2.0 Client ID
>Type: Chrome Extension (Chrome) — add your extension ID
>Or: Web application (cross-browser) — add chrome.identity.getRedirectURL() as an authorised redirect URI
>Copy the client ID into MALSync → Settings → Backup & Restore → Google Drive → OAuth Client ID
>
>The drive.appdata scope means the consent screen says "See and download your application-specific data in Google Drive" no access to the user's actual files.


lolamtisch:
Is the complexity of backuping to the cloud really necessary? It is a file you download once after all.
It seems to include much more then the settings and as far I see it has no correct handling of the settings during import either. I am pretty sure the import in general will fail too.
Including credententials is a very bad idea. The backups will land in the public no matter what. And it most cases it is not that useful, because credentials get rotateted regulary.

Ash1421:
> Is the complexity of backuping to the cloud really necessary? It is a file you download once after all. It seems to include much more then the settings and as far I see it has no correct handling of the settings during import either. I am pretty sure the import in general will fail too. Including credententials is a very bad idea. The backups will land in the public no matter what. And it most cases it is not that useful, because credentials get rotateted regulary.

As I have said I have done this while I'll still I'll and your welcome to fix whatever you don't like, if someone changes stuff then they would have to redownload eg diff theme then they have to redownload/sync file (automated sync would be good) didn't really think about auto syncing, yeah kind of thought that about creds and i know the handling isn't great.

lolamtisch:
Malsync already syncs settings between devices so that is not really a concern. The commit did not fix any of my other concerns.

Ash1421:
> Malsync already syncs settings between devices so that is not really a concern. The commit did not fix any of my other concerns.

Yeah maybe, for like Firefox logged in or something but how else ???

and what are your other concerns then.

PR #4173, opened 2026-08-31
[New Website] Add flixmomo.app

kennyL07:
Add flixmomo.app.


lolamtisch (changes requested):
You will have to make sure that it only tracks anime. Currently it tracks western series too.

kennyL07:
Is there a way to get the MAL results from MALSync? It doesn't seem like I can make an api request to Tenrai. I was thinking about recreating this as a legacy implementation instead.

lolamtisch:
You could look at the Plex/jellyfin implementations. While it is not possible to do requests, it is possible to read requests the website does. 

We do not really want the old implementations, as you will only be able to update the implementation every 2-6 months

kennyL07:
I'm not sure if this is the intended way to read requests, but it seems like it works. Does it matter if I filter only the TMDB request or just use all requests that go through to check for metadata?

src/pages-chibi/implementations/FlixMomo/main.ts
```diff
@@ -0,0 +1,174 @@
...
+    getIdentifier($c) {
+      return getIdentifier($c).run();
+    },
+    getImage($c) {
+      return $c.querySelector('main img.poster-img').getAttribute('src').ifNotReturn().run();
+    },
+    uiInjection($c) {
+      return $c.querySelector('main h4').uiAfter().run();
+    },
+  },
```
lolamtisch: I think overview support makes no sense, as there is no possiblity to select a season. Probably better to drop it

lolamtisch (changes requested):

src/pages-chibi/implementations/FlixMomo/main.ts
kennyL07: I didn't notice that there weren't individual pages for seasons. mb

lolamtisch (approved):

PR #4191, opened 2026-09-06
chore: fr page maintenance

NatsuDzn:
Removed Bentomanga because it doesn't exist anymore


lolamtisch:
We dont accept changes to leagacy implementations anymore. You will need to migrate it to a chibi implementation.

NatsuDzn:
> We dont accept changes to leagacy implementations anymore. You will need to migrate it to a chibi implementation.

It should be fine now

lolamtisch (approved):

